PR #5, opened 2015-03-04
Add declaration_keyword configuration option

lencioni:
If you are using ES6 (ES 2015), you have access to `let` and `const` in
addition to `var` as ways to declare variables. `const` seems like an
appropriate declaration keyword to use when requiring modules, so it
would be nice if this plugin supports it.

This commit adds support for a `declaration_keyword` option that can be
set to a string that will be used instead of `var` when declaring the
variables that are assigned to the required modules.

### DIFF
--- a/spec/import-js/importer_spec.rb
+++ b/spec/import-js/importer_spec.rb
@@ -273,6 +273,31 @@ $
         EOS
         end
       end
+
+      context 'with declaration_keyword' do
+        subject do
+          ImportJS::Importer.new.import
+          VIM::Buffer.current_buffer.to_s
+        end
+
+        let(:configuration) do
+          {
+            'declaration_keyword' => 'const'
+          }
+        end
+
+        context 'with a variable name that will resolve' do
+          let(:resolved_files) { ['bar/foo.js.jsx'] }
+
+          it 'adds an import to the top of the buffer using the declaration_keyword' do
+            expect(subject).to eq(<<-EOS.strip)
+const foo = require('bar/foo');
+
+foo
+            EOS
+          end
+        end
+      end
     end
   end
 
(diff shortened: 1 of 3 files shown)


lencioni:
It looks like the Travis build failed on an unrelated change in an earlier commit on master, only because you are using the Ruby 1.9 style and running tests on Ruby 1.8.7 (which was EOL'd in June 2013). What do you think about removing 1.8.7 from the test runs?

lencioni:
This isn't quite right yet I don't think, since there are some other places that depend on `var`. Hopefully I'll have some time soon to fix it up.

lencioni:
@trotzig Alright, I think this is ready to go now. I expect the Travis build to fail unrelatedly until #7 is merged.

spec/import-js/importer_spec.rb
```diff
@@ -273,6 +273,66 @@ def self.current_selection=(index)
...
+let bar = require('foo/bar');
+
+foo
+            EOS
+
+            it 'adds the import and sorts the entire list' do
+              expect(subject).to eq(<<-EOS.strip)
+const foo = require('bar/foo');
+let bar = require('foo/bar');
+var zoo = require('foo/zoo');
```
trotzig: Since foo, bar, zoo comes in the same order as const, let, var; would it make sense to shuffle the variable names a bit so that it's clearer what is used for sorting (variable name only or declaration keyword + variable name)?

lencioni: I think that's exactly what I did, since alphabetizing the variable names would result in "bar, foo, zoo" no? Or maybe I don't understand your suggestion.

trotzig: I'm jet lagged. You are right.

PR #6, opened 2015-03-08
Add .projections.json

lencioni:
This configuration help plugins like vim-projectionist work. With this
file, I can now run `:A` in Vim to jump between the Ruby source file and the
spec file, making it easier to work with this project.


trotzig:
This isn't super-important in a project of this size, but I can't see the harm of having it here.

PR #7, opened 2015-03-08
Remove Ruby 1.8.7 from Travis, add 2.2.1

lencioni:
This project uses Ruby syntax that was introduced in 1.9.0. Furthermore,
Ruby 1.8.7 was EOL'd in June 2013. It doesn't seem valuable to run tests
against this version of Ruby anymore.

https://www.ruby-lang.org/en/news/2013/06/30/we-retire-1-8-7/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: ruby
 rvm:
-  - "1.8.7"
   - "1.9.3"
   - "2.1.1"
+  - "2.2.1"
 
 script: rspec
 install: gem install rspec


trotzig:
Before I merge this, I just want to make sure that @paradasia thinks it's a good idea. She added support for pre 1.9 ruby in https://github.com/trotzig/import-js/commit/e17b3d47781da5007d106ee05d1071034f309b88, and I remember this having to do with the default Ruby version in OSX (I might be wrong).

Do you know what commit broke it?

lencioni:
If you look at the [build history](https://travis-ci.org/trotzig/import-js/builds), it appears to be something between 6c910e14813a96297e1ff0232fd7cc44d73ef0c6 and  c07feb27ca9e11a74d73f29ac635d978876cb066.

https://github.com/trotzig/import-js/compare/6c910e14813a96297e1ff0232fd7cc44d73ef0c6...c07feb27ca9e11a74d73f29ac635d978876cb066

I think it is probably the commit that added this line: https://github.com/trotzig/import-js/compare/6c910e14813a96297e1ff0232fd7cc44d73ef0c6...c07feb27ca9e11a74d73f29ac635d978876cb066#diff-2bed0707fdd562ca8d11a085926cfcc7R37

lencioni:
Also, Yosemite ships with Ruby 2.0.0 by default, but the previous version shipped with 1.8.7.

trotzig:
Ah, my bad then. :)

I'm merging this even though I haven't heard back from @flarnie. We can always revert/fixup later.

Issue #9, opened 2015-03-19
Does not understand line breaks


lencioni:
If I have a section of `require`d modules at the top that looks like this:

```js
const MyModule = require('path/to/my_module');
const MyModuleThatHasALongName =
  require('path/to/my_module_that_has_a_long_name');
const SomeModule = require('path/to/some_module');
```

import-js gets confused when adding a new module to the list:

```js
const AnotherModule = require('path/to/another_module');
const MyModule = require('path/to/my_module');

const MyModuleThatHasALongName =
  require('path/to/my_module_that_has_a_long_name');
const SomeModule = require('path/to/some_module');
```

where I would expect:

```js
const AnotherModule = require('path/to/another_module');
const MyModule = require('path/to/my_module');
const MyModuleThatHasALongName =
  require('path/to/my_module_that_has_a_long_name');
const SomeModule = require('path/to/some_module');
```

It would be nice if import-js worked better with this format.

Bonus points would be specifying a character limit (defaulted to `80`) that it would try to stay within when importing new dependencies.

trotzig:
I made a change to deal with line-breaks better. I didn't add anything for auto-linebreaking yet, but I might do that next. Check it out and let me know what you think @lencioni!

lencioni:
:+1: :rabbit: :tophat:

lencioni:
I think this is closed now. I'll open a new issue for the auto-breaking behavior.

PR #10, opened 2015-03-21
Add support for imports with line-breaks

trotzig:
As reported by @lencioni in #9:

  If I have a section of required modules at the top that looks like this:

    const MyModule = require('path/to/my_module');
    const MyModuleThatHasALongName =
      require('path/to/my_module_that_has_a_long_name');
    const SomeModule = require('path/to/some_module');

  import-js gets confused when adding a new module to the list:

    const AnotherModule = require('path/to/another_module');
    const MyModule = require('path/to/my_module');

    const MyModuleThatHasALongName =
      require('path/to/my_module_that_has_a_long_name');
    const SomeModule = require('path/to/some_module');

  where I would expect:

    const AnotherModule = require('path/to/another_module');
    const MyModule = require('path/to/my_module');
    const MyModuleThatHasALongName =
      require('path/to/my_module_that_has_a_long_name');
    const SomeModule = require('path/to/some_module');

  It would be nice if import-js worked better with this format.

I've added basic support for line breaks in this commit. I say basic
because I expect there to be edge cases that I haven't thought of.
We use simple regexs to parse the js file content, and like all simple
regex parsers, they are bound to fail. However, I don't want to bring in
a js parsing framework just yet. So in this commit I extend the regex
approach with some logic to deal with possible line-breaks.

One idea also listed in #9 was to add support for import-js to
automatically line-break long lines. I haven't made that change in this
commit (new imports are still one-line), but I might do that later if I
find time.

### DIFF
--- a/ruby/import-js/importer.rb
+++ b/ruby/import-js/importer.rb
@@ -103,42 +103,55 @@ module ImportJS
     # @return [number] the number of lines changed
     def write_imports(variable_name, path_to_file)
       current_imports = find_current_imports
-      before_length = current_imports.length
-      current_imports.length.times do
+
+      unless buffer[current_imports[:newline_count] + 1].strip.empty?
+        # Add a newline after imports
+        buffer.append(current_imports[:newline_count], '')
+      end
+
+      # delete current imports
+      current_imports[:newline_count].times do
         buffer.delete(1)
       end
 
-      declaration_keyword = @config['declaration_keyword']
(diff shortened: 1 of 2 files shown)


trotzig:
I'm going to merge this as is. @lencioni those are all valid comments but I don't have time to address right now. Thanks for taking the time to review!

lencioni:
Sounds good. I'll send you a PR with these changes then.

PR #11, opened 2015-03-23
Improve readability of write_imports

lencioni:
As discussed in https://github.com/trotzig/import-js/pull/10, we weren't
very happy with how this method reads, so we decided to apply some
changes to make it more understandable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,14 @@ Configure a path to a `jshint` compatible command, e.g. `jsxhint`.
 jshint_cmd: jsxhint
 ```
 
+### `text_width`
+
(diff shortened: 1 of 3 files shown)


lencioni:
I went ahead and did the auto-breaking here too.

lencioni:
I think I have this done now, but I didn't actually try out my latest commit. I gotta run, so if you want this merged soon and there are any problems, will you take care of them?

trotzig:
Yep! There might still be an issue with moving the cursor when adding multilined imports, but that's a minor thing.

PR #20, opened 2015-04-26
Improve ESLint compatibility

lencioni:
We are using a JSHint configuration comment to ensure that the settings
we need to find unused variables is enabled, but we weren't doing
anything similar for folks who are using ESLint. To make sure that this
will work regardless of a project's ESLint configuration, I am adding a
similar config comment to enable the rule that we are relying on for
this feature. I disabled it checking args for this because those don't
matter for this use case and I assume that disabling it will make it
faster.

I wonder if we can get some more performance by disabling all rules but
the one that we need, but that will have to wait for another time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ to `var` as ways to declare variables. To use one of these, set the
 
 ### `jshint_cmd`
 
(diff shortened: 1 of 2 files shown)


trotzig:
Sorry, missed this for some reason. Looks good.

PR #25, opened 2015-06-08
Allow matching dashed and camel cased filenames; fixes #24

### LINKED ISSUE #24: Match dashed and camel cased filenames
It seems like `import-js` doesn't find files if they use dashed or camel-cased filenames. My first thought is that by default it should look for all matching filenames that are either snake-cased, dashed, or camel cased, and users can configure it to only use one if they prefer. That way the default works for everyone, no configuration required.

The alternative would be to pick one as the default and allow users to configure it to use one or more of the other patterns.

flarnie:
There is not a clear consensus on best practices for naming file in
JavaScript, but generally one of three conventions is used:[1]
 - dashed (as in `my-javascript-file.js`)
 - snake case (as in `my_javascript_file.js`)
 - camel cased (as in `MyJavaScriptFile.js`)

Most projects I have seen use either dashed or snake case, but React.js
(last I checked) does use camel case and I know other projects might as
well.

To be useful for more folks and in more projects, `import-js` should
match based on any of the three naming conventions commonly used in
JavaScript.

For the future, there should be tests written to cover this feature, and
an API for changing the file name matching algorithm via configuration.

[1]: http://stackoverflow.com/questions/7273316/what-is-the-javascript-filename-naming-convention

### DIFF
--- a/ruby/import_js/importer.rb
+++ b/ruby/import_js/importer.rb
@@ -164,11 +164,13 @@ module ImportJS
       if alias_path = @config.get('aliases')[variable_name]
         return [alias_path]
       end
-      snake_case_variable = camelcase_to_snakecase(variable_name)
+      regex_variable = formatted_to_regex(variable_name)
       matched_files = []
       @config.get('lookup_paths').each do |lookup_path|
         Dir.chdir(lookup_path) do
-          matched_files.concat(Dir.glob("**/#{snake_case_variable}.js*"))
+          matched_files.
+            concat(Dir.glob("**/#{regex_variable}.js*", File::FNM_CASEFOLD).
+            select { |element| element.match(/#{regex_variable.gsub('*', '.?')}/i) })
         end
       end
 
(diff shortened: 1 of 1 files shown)


ruby/import_js/importer.rb
```diff
@@ -164,11 +164,18 @@ def find_files(variable_name)
...
       matched_files = []
       @config.get('lookup_paths').each do |lookup_path|
         Dir.chdir(lookup_path) do
-          matched_files.concat(Dir.glob("**/#{snake_case_variable}.js*"))
+          matched_files.concat(
+            Dir.glob([
+              "**/#{dashed_variable}.js*",
+              "**/#{snake_case_variable}.js*",
+              "**/#{variable_name}.js*"
+            ])
```
trotzig: I ran some benchmarking to see the performance penalty of introducing three matchers to `glob`, and it looks like the perf penalty is linear, at least on a codebase of Brigade's size. I've already had some complaints about performance from @lencioni before (https://github.com/trotzig/import-js/issues/22), and I'm a little hesitant to merge this as-is. I totally understand and agree with the rationale behind this commit though, so I'll list a few options I can think of to have the same functionality without introducing too much performance debt.

1. Make filename convention configurable so that only one glob pattern has to be used.
2. Reduce the three matchers to two by using the `?` matcher. E.g. `Dir.glob(['**/icon?button.js*', '**/IconButton.js*'])`
3. Reduce the three matchers to one: `Dir.glob('**/icon*button.js*'). There's a targeting penalty with that though, as e.g. `/js/components/icon_alternative_button.jsx` will also be matched. That might not be a big deal though.

Here's a benchmark I ran:
```
irb(main):022:0> puts Benchmark.measure { Dir.glob('**/icon_button.js*') }
  0.040000   0.150000   0.190000 (  0.190436)
=> nil
irb(main):023:0> puts Benchmark.measure { Dir.glob(['**/icon_button.js*', '**/IconButton.js*', '**/icon-button.js*']) }
  0.110000   0.450000   0.560000 (  0.568228)
=> nil
```

flarnie: Those are good ideas - thanks! I will fix this up and resubmit.

flarnie: I'm about to push a new version - the '**/icon*button.js*' has the best perf., still need to do some QA to see how well it works.

trotzig:
Looks good! Thanks for taking the time to fix this! I'll merge as-is and let you follow up with changes if you want to.

PR #28, opened 2015-07-09
Match modules following node.js naming conventions

trotzig:
At Brigade, we've started using a component-centric structure for our
React code base. This means that everything that belongs to a component
(scss code, tests, jsx code) is stored in one folder. E.g.

-- Button
   |
   |-- index.jsx
   |-- index.scss
   |-- index-test.jsx

We want ImportJS to find the index.jsx files, and import it in the
following way:

  const Button = require('<path_to_modules>/Button');

I made a first naive implementation by adding another Dir.glob, looking
for anything with the module name and ending in index.js*. The previous
post-processing filter was enough to rule out any mismatched files.

I understand that this introduces a performance penalty. I intend to
follow up by making the Dir.glob call even more broad, and let the
post-processing filter take care of reducing even further.

### DIFF
--- a/ruby/import_js/importer.rb
+++ b/ruby/import_js/importer.rb
@@ -168,16 +168,14 @@ module ImportJS
       matched_files = []
       @config.get('lookup_paths').each do |lookup_path|
         Dir.chdir(lookup_path) do
-          matched_files.
-            # Look for files named similar to the variable
-            concat(Dir.glob("**/#{regex_variable}.js*", File::FNM_CASEFOLD)).
-            # Look for modules following the node.js convention of
-            # module_name/index.js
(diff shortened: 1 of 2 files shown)


trotzig:
Thanks @lencioni. I saw that issue too, and I was going to tackle that after this one. Though that's a bit harder and I didn't want to complicate this fix too much.

I'll amend this shortly.

trotzig:
I hit some road bumps along the way and will have to continue tomorrow.

lencioni:
No worries. Company offsite tomorrow so people probably won't benefit from these changes until Monday anyway. Thanks for working on this! import-js has become a very useful tool for us. I think that @bigsley is thinking about doing something similar for Sass.

trotzig:
Cool, I'm glad you find it useful.

PR #30, opened 2015-07-14
Import cache

trotzig:
@lencioni, @flarnie et al: If you have time, could you try this branch out to see if performance has improved? I added a "smart" cache so that we don't have to scan the filesystem every time you import.


trotzig:
I decided to merge it anyway. I'll follow up if it turns out that it doesn't improve anything.

PR #45, opened 2015-07-28
Auto-reload configuration

### LINKED ISSUE #44: Feature request: auto-reload config without restarting Vim
I find that whenever I modify the `.importjs.json` file, I need to restart Vim before the new settings take effect. It would be nice if I the settings were just reloaded without needing to restart Vim.

lencioni:
I find that whenever I modify the .importjs.json file, I need to restart
Vim before the new settings take effect. It would be nice if I the
settings were just reloaded without needing to restart Vim.

Fixes #44

### DIFF
--- a/ruby/import_js/configuration.rb
+++ b/ruby/import_js/configuration.rb
@@ -12,9 +12,14 @@ module ImportJS
     'lookup_paths' => ['.'],
   }
 
-  # Class that initializes configuration from a .importjs file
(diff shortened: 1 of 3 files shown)


trotzig:
Looks good, I'm going to see if I can fix the failing spec before merging.

lencioni:
I pushed a new version that fixes the spec (and adds a spec for this new behavior).

trotzig:
Looks good!

PR #49, opened 2015-11-08
Dedupe modules when lookup_paths overlap

trotzig:
At Brigade, we have two lookup_paths that overlap:

  app/assets
  app/assets/javascripts

When you try to import a module in `app/assets/javascripts`, you get two
options:

  [import-js] Pick js module to import for 'warn': (0.24s)
  1: javascripts/lib/warn
  2: lib/warn

The overlapping lookup paths are for many reasons not ideal, but
changing this structure isn't a straightforward change. So I've decided
to do something about it in import-js instead.

Assuming that we always want the shorter import path (`lib/warn` in the
example above), we first sort modules by import path length, then dedupe
modules. This will remove the module(s) with the longer path(s). We then
sort again, to get consistent alphabetic sorting in the selection list.

### DIFF
--- a/ruby/import_js/importer.rb
+++ b/ruby/import_js/importer.rb
@@ -210,7 +210,15 @@ module ImportJS
           end.compact
         )
       end
-      matched_modules.uniq { |m| m.import_path }.sort do |a, b|
+
+      # If you have overlapping lookup paths, you might end up seeing the same
+      # module to import twice. In order to dedupe these, we remove the module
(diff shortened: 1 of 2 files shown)


trotzig:
For some reason, the `git clone` didn't work here: https://travis-ci.org/trotzig/import-js/jobs/89975380

The checks for Ruby 1.9.3 and 2.2.1 passed, so I'm merging anyway.

PR #53, opened 2015-11-11
Truncate messages to prevent having to confirm

### LINKED ISSUE #48: Truncate logs to prevent having to confirm
Sometimes, when you're importing a module with a long name/path to, the Vim buffer width isn't big enough to display the full message. In those cases, you have to hit Enter in order to continue doing what you were doing in the file. We could fix this by looking at the width of the buffer and automatically truncating the log message.

lencioni:
Sometimes, when you're importing a module with a long name/path to, Vim
isn't wide enough to display the full message. In those cases, you have
to hit Enter in order to continue doing what you were doing in the file.

In this commit I fix this by looking at the width of Vim and
automatically truncating any messages being displayed.

Fixes #48

### DIFF
--- a/autoload/importjs.vim
+++ b/autoload/importjs.vim
@@ -8,6 +8,17 @@ function importjs#ImportJSGoTo()
   ruby $import_js.goto
 endfunction
 
+" WideMsg() prints [long] message up to (&columns-1) length
+" guaranteed without "Press Enter" prompt.
+" http://vim.wikia.com/wiki/How_to_print_full_screen_width_messages
+function! importjs#WideMsg(msg)
+  let x=&ruler | let y=&showcmd
+  set noruler noshowcmd
+  redraw
+  echo a:msg
+  let &ruler=x | let &showcmd=y
+endfun
+
 ruby << EOF
   begin
     require 'import_js'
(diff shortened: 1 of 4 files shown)


trotzig:
Nice! Looks like you have to make a small adjustment in order to get the Ruby 1.9.3 build to pass.

lencioni:
I pushed a new version. Hopefully that fixes it.

lencioni:
Specs are passing. I'm trying this out and unfortunately it doesn't seem to prevent the press enter to continue.

lencioni:
Looking at `:help hit-enter`:

```
This message is given when there is something on the screen for you to read,
and the screen is about to be redrawn:
- After executing an external command (e.g., ":!ls" and "=").
- Something is displayed on the status line that is longer than the width of
  the window, or runs into the 'showcmd' or 'ruler' output.
```

So there might be additional work that needs to be done here.

lencioni:
This looks like a useful resource: http://vim.wikia.com/wiki/How_to_print_full_screen_width_messages

trotzig:
Maybe you just need to remove one more character? 

Also, have you seen the tips for developing locally here? https://github.com/trotzig/import-js/blob/master/CONTRIBUTING.md

lencioni:
I do need to remove one more character, but there are other things in the way. I just pushed a new version that should be good to go.

PR #57, opened 2015-11-16
Command line

trotzig:


trotzig:
@lencioni any ideas why the ellipsize character is messing up the build in Ruby 1.9.3?

trotzig:
Nevermind, I found out how to work around it:
http://stackoverflow.com/questions/3678172/ruby-1-9-invalid-multibyte-char-us-ascii

lencioni:
Yep, you need to add `# encoding: utf-8` to the top of the file for Ruby 1.9.3.

Alternatively, you could choose to not support this version, since it was EOLd in February.

trotzig:
True. I think we added it to make sure that we supported as many VIM environments as possible. And I haven't really felt any huge pains from it, so we might as well keep it around.

PR #59, opened 2015-11-16
Use %q{...} instead of "..." for eslint config comment

lencioni:
This prevents us from having to escape all of the double quotes, which
makes this line a bit easier to read and work with.

### DIFF
--- a/lib/import_js/importer.rb
+++ b/lib/import_js/importer.rb
@@ -24,7 +24,8 @@ module ImportJS
 
       old_buffer_lines = @editor.count_lines
       import_one_variable variable_name
-      return unless lines_changed = @editor.count_lines - old_buffer_lines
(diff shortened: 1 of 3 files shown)


lib/import_js/importer.rb
```diff
@@ -249,7 +249,7 @@ def resolve_one_js_module(js_modules, variable_name)
 
       selected_index = @editor.ask_for_selection(
         "\"[import-js] Pick js module to import for '#{variable_name}': #{timing}\"",
-        js_modules.map {|m| m.display_name}
```
trotzig: thanks :)

lencioni:
I fixed merge conflicts and added another commit on top

trotzig:
Cool, thanks for the cleanup!

PR #61, opened 2015-11-22
Add basic Emacs support

kevinkehl:
This has a long way to go to become a full featured setup for import-js,
but it is functional.

Adds two pieces:
1. ImportJS::EmacsEditor class that uses the ImportJS APIs to resolve
modules and add `require` lines.
2. An elisp plugin to start up and talk to a Ruby process that runs the
ImportJS::EmacsEditor.

For now, this works a little differently from the Vim plugin. We don't
have direct access to the Ruby code in our buffers; instead, we
communicate with the process using stdin/stdout, and leave the file
handling to Ruby.

I think there's some additional optimizations to be made by further
breaking out the core import functionality from the original Vim plugin,
and also implementing the 'experimental' goto interface.

I'd love some help/thoughts on what specs should be included.

### DIFF
--- a/EMACS.md
+++ b/EMACS.md
@@ -1,0 +1,24 @@
+# Running import-js in Emacs
+
+1. Install import-js
+  * `gem install import-js`
+2. Configure import-js
+  * See [Configuration](README.md#configuration)
+2. Install import-js.el for Emacs
+  * This is not yet available on elpa/melpa, so for now this means dropping
+    import-js.el somewhere in your .emacs.d
+3. Configure your project root
+  * `(setq import-js-project-root "/path/to/project")`
+4. Start the import js project
+  * `(M-x) run-import-js`
+5. Import a file!
+  * You can use something like `(M-x) import-js-import` with your cursor over
+    the desired module
+  * It will be helpful to bind `import-js-import` to an easy-to-use binding,
+    such as:
+
+    ```
+    (define-prefix-command 'my-keymap)
+    (global-set-key (kbd "s-a") 'my-keymap)
+    (define-key my-keymap (kbd "a u") 'import-js-import)
+    ```
--- a/lib/import_js/emacs_editor.rb
+++ b/lib/import_js/emacs_editor.rb
@@ -1,0 +1,145 @@
+# encoding: utf-8
+module ImportJS
+end
+
+class ImportJS::EmacsEditor
+  attr_accessor :current_word
+
+  def initialize
+    loop do
+      input = gets.chomp
+      command, value, path = input.split(':')
+
+      begin
+        @path = path
+        @file = File.readlines(path).map(&:chomp)
+
+        case command
+        when 'import'
+          @current_word = value
+          ImportJS::Importer.new(self).import
+          write_file
+          puts 'import:success'
+        when 'goto'
+          ImportJS::Importer.new(self).goto
+        end
+      rescue Exception => e
+        puts e.inspect
+      end
+    end
+  end
+
+  def write_file
+    new_file = File.open(@path, 'w')
+    @file.each {|line| new_file.puts(line) }
+    new_file.close
+  end
+
+  # Open a file specified by a path.
+  #
+  # @param file_path [String]
+  def open_file(file_path)
+    puts file_path
+  end
+
+  # Display a message to the user.
+  #
+  # @param str [String]
+  def message(str)
+    puts str
+  end
+
+  # Read the entire file into a string.
+  #
+  # @return [String]
+  def current_file_content
+    @file.join('\n')
+  end
+
+  # Reads a line from the file.
+  #
+  # Lines are one-indexed, so 1 means the first line in the file.
+  # @return [String]
+  def read_line(line_number)
+    @file[line_number - 1]
+  end
+
+  # Get the cursor position.
+  #
+  # @return [Array(Number, Number)]
+  def cursor
+    return 0, 0
+  end
+
+  # Place the cursor at a specified position.
+  #
+  # @param position_tuple [Array(Number, Number)] the row and column to place
+  #   the cursor at.
+  def cursor=(position_tuple)
+  end
+
+  # Delete a line.
+  #
+  # @param line_number [Number] One-indexed line number.
+  #   1 is the first line in the file.
+  def delete_line(line_number)
+    @file.delete_at(line_number - 1)
+  end
+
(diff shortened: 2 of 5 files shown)


lib/import_js/emacs_editor.rb
```diff
@@ -0,0 +1,145 @@
+# encoding: utf-8
+module ImportJS
+end
```
trotzig: It depends on how you load this class, but there's some code in `lib/import_js.rb` that should make sure the module is never empty.

lib/import_js/emacs_editor.rb
```diff
@@ -0,0 +1,145 @@
...
+  # @param str [String]
+  def message(str)
+    puts str
+  end
+
+  # Read the entire file into a string.
+  #
+  # @return [String]
+  def current_file_content
+    @file.join('\n')
```
trotzig: I forget where this is used, but don't you need to surround this with double quotes? `"\n"`

trotzig:
Thanks @kevinkehl!

In our current specs, VIM is completely mocked out. I'd actually be fine letting this be manually tested until we have a better sense for the general direction.

I'll give this a spin myself just to make sure that it doesn't interfere with the VIM plugin.

EMACS.md
```diff
@@ -0,0 +1,24 @@
+# Running import-js in Emacs
+
+1. Install import-js
+  * `gem install import-js`
+2. Configure import-js
+  * See [Configuration](README.md#configuration)
+2. Install import-js.el for Emacs
+  * This is not yet available on elpa/melpa, so for now this means dropping
+    import-js.el somewhere in your .emacs.d
```
trotzig: I'm assuming you mean that the file should be grabbed from within the gem? Or did you mean grab it from the repo directly?

trotzig:
I'm running into some issues when trying this out locally. I'll go through what I did to get it setup:

1. checked out your fork of the code
2. changed version to 0.0.3 and ran `gem build import_js.gemspec; gem install --local import_js-0.0.3.gem`
3. copied the contents of `import-js.el` into `~/.emacs.d/init.el` (I first tried just copying the file into the `~/.emacs.d/` folder, but that didn't work.
4. Put `(setq import-js-project-root "~/brigade")` at the top of  `~/.emacs.d/init.el`
5. fired up `emacs app/assets/components/Button/index.jsx`
6. ran `M-x run-import-js`, which gave me output `Symbol's function definition is void: split-string-and-unquote`
7. put my cursor on `Icon`.
8. ran `M-x import-js-import`, editor asks me `Save file Users/henrictrotzig/brigade/app/assets/components/Button/index.jsx? (y, n, !, ., q, C-r, d or C-h)`
9. I hit `y`
10. nothing was imported

Where do you think I'm failing? Bare with me, I haven't used emacs since 2004...

trotzig:
Hm, I thought my emacs install worked, but it looks like it didn't. I'm on `22.1.1` but I see that `24.5` is the latest. I'll try again.

trotzig:
Okay, I'm on `24.5` now. I'm still unable to import anything. Now I get `Buffer *import-js* has no process` when I run `M-x import-js-import`.

lib/import_js/emacs_editor.rb
kevinkehl: Actually probably yes, I'm not sure why this works...

EMACS.md
kevinkehl: Well right now its not being packaged in the gem, so the only place to get it is the repo. The much better way to do this is publish it on melpa, but also want it stable first.

kevinkehl:
Oh yea, this will probably only work as stated on Emacs 24, I should specify that.

It looks like the process is failing to start up, so importing won't work. 

Could you do me a favor and do (M-x) describe-function <ret> split-string-and-unquote <ret> ? This should tell us if that function should be available.

Dropping the file in the directory doesn't work without requiring it in your init.el, I will specify that.

kevinkehl:
Also, is there any output in the *import-js* buffer?

kevinkehl:
Also again, I am setting the project root in a mode hook like so: 

```
(defun setup-web-mode ()
  ;; start the import-js process
  (setq import-js-project-root "/my/project/path")
)

(add-hook 'web-mode-hook 'setup-web-mode)

```

Before running `run-import-js`. Making it run correctly on entering the mode is on my list, but not necessary for the first go.

trotzig:
`M-x describe-function split-string-and-unquote` gives me 
```
split-string-and-unquote is a compiled Lisp function in `subr.el'.

(split-string-and-unquote STRING &optional SEPARATOR)

Split the STRING into a list of strings.
It understands Emacs Lisp quoting within STRING, such that
  (split-string-and-unquote (combine-and-quote-strings strs)) == strs
The SEPARATOR regexp defaults to "\s-+".
```

The `*import-js*` buffer has an error which should be easy to fix!
```
File Edit Options Buffers Tools Complete In/Out Signals Help
-e:1:in `chdir': No such file or directory @ dir_chdir - ~/brigade (Errno::ENOENT)
        from -e:1:in `<main>'

Process import-js exited abnormally with code 1
```

trotzig:
Hey, success!
```
const Icon = require('components/Icon');

Icon
```

I changed the project root path so that it wasn't using the tilde `~`.

Though I also get this down at the bottom:
`error in process filter: Wrong type argument: processp, nil`

trotzig:
Do you want to continue working on this or should we merge and then continue in other PRs?

kevinkehl:
I'd like to first figure out what that error you are getting is from, make sure it won't cause horrible issues. Did the buffer refresh for you so you see the import automatically?

trotzig:
I think it did. I'm not on my computer atm, but I'll give it a try tomorrow.

trotzig:
I couldn't catch all the messages flashing by, so I made a gif.

emacs-import-js

kevinkehl:
Thanks for the gif. I managed to reproduce and get the same warning. It doesn't seem to be affecting the behavior though.

By the way, you can see all the messages that had been displayed in the minibuffer by changing to the  *Messages* buffer.

trotzig:
I'll merge this and update the README a bit. I think I'll mention that the Emacs support is experimental since we haven't worked out all the kinks.

kevinkehl:
Thanks! I got caught up in other things but plan to also try to work out a few things this weekend, at least enough to clear any error messages.

PR #65, opened 2015-12-03
Fix a couple of small bugs when deduping

lencioni:
I noticed a couple of small bugs in the deduping code.

1. If the module contained "from" in the name, the deduping could break.
2. If the import was using double quotes, the deduping could break.

This PR has commits that fix both of these problems, each with a test that failed before the fix.

### DIFF
--- a/lib/import_js/importer.rb
+++ b/lib/import_js/importer.rb
@@ -103,7 +103,9 @@ module ImportJS
         (.+?)                # \1 variable assignment
         \s*=\s*
         require\(
-          ('.*')             # \2 module path
+          ('|")              # \2 opening quote
+          ([^\2]+)           # \3 module path
+          \2                 # closing quote
         \);?
         \s*
       }xm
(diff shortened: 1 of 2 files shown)


lib/import_js/importer.rb
```diff
@@ -93,6 +93,39 @@ def import_one_variable(variable_name)
...
+        ('|")              # \2 opening quote
+        ([^\2]+)           # \3 module path
+        \2                 # closing quote
+        ;?\s*
+      }xm
+
+      import
+        .sub(const_let_var_regex, '\1 \3')
+        .sub(import_regex, '\1 \3')
+    end
```
trotzig: I wonder if we can double-use this regex to also find imports?

trotzig: BTW: this is soooooo much better than what was there before!

trotzig:
Thank you!

lib/import_js/importer.rb
trotzig: Regarding the double-use: I'm going to try that out myself.

lencioni: Sounds good. If you plan on moving these regexes farther away from where they are used, you might consider using named capture groups instead of numbers.

lencioni: I switched to named capture groups in 684f1d4cd8.

PR #66, opened 2015-12-03
DRY up regexes

lencioni:


trotzig:
I might follow up with using these for the destructured stuff as well. Probably by moving the regex and parsing into its own ImportJS::ImportStatement class. We can then do stuff like

```
statement = ImportJS::ImportStatement.parse(import_line)
return unless statement # did not match
statement.normalize # for the uniq! comparison
statement.destructured? # to check if the assignment contains destructured variables
statement.destructured_variables # to get the destructured list
statement.to_s # to reconstruct the statement (with line-breaking etc)
```

That way we can abstract away the declaration_keyword etc.

trotzig:
Here's that idea implemented: https://github.com/trotzig/import-js/pull/68

PR #70, opened 2015-12-08
Correctly save current buffer before reverting and add goto

kevinkehl:
Since we are switching the buffer context in order to revert, we need to
properly handle switching back so other functions operate normally.

This clears the 'error in process filter: Wrong type argument: processp, nil'
error we were seeing in:

https://github.com/trotzig/import-js/pull/61

Also, implements goto:

Make goto work with Emacs

While experimental and not necessarily bound to import-js, implementing
goto using the same underlying package finding logic is very simple and
extremely useful.
    
Implement the 'goto' functionality with Emacs, so you can easily
navigate to packages with a simple keystroke.

### DIFF
--- a/EMACS.md
+++ b/EMACS.md
@@ -22,3 +22,8 @@
     (global-set-key (kbd "s-a") 'my-keymap)
     (define-key my-keymap (kbd "a u") 'import-js-import)
     ```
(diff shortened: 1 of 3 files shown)


trotzig:
+1

PR #72, opened 2015-12-14
Fix up ESLint integration

lencioni:
I found a few improvements to make that improves my experience when using the fix imports functionality of ImportJS.


trotzig:
lgtm

PR #73, opened 2015-12-14
Fix .projections.json

lencioni:
This file was using dashes where we actually should be using
underscores.


trotzig:
Ah, I noticed that this didn't work. Thanks for fixing!

PR #78, opened 2015-12-22
Add ability to use `{filename}` in aliases

### LINKED ISSUE #76: Dynamic aliases
I am working in an application components are not structured into directories that contain `index.jsx`, `index.scss`, etc. In this structure, components are in files like `MyComponent.jsx` and their styles are in files like `MyComponent.scss`. In this case, it would be nice to be able to make a dynamic alias for `styles` that resolves to the current filename plus `.scss`.

trotzig:
By using {filename} in an alias, you can use the name of the currently
edited file in the resulting import statement. @lencioni suggested
looking at vim-projectionist [1] for inspiration. Which I did, but I
decided to not use much of it. The main reason being that projecting is
a different usecase, where you have configuration that apply on a large
set of files. We want to apply a configuration on a single alias.

To keep it simple, I only added the filename (without file extension) to
begin with. I can imagine usecases where you need more than just that
though (e.g. a relative path within the project, or a relative path from
a `lookup_path`). We can extend this functionality when those usecases
pop up.

[Fixes #76]

[1]: https://github.com/tpope/vim-projectionist

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,23 @@ const { memoize } = require('underscore');
 memoize(() => { foo() });
 ```
 
+Aliases can be made dynamic by using the `{filename}` string. This part of the
+alias will be replaced by the name of the file you are currently editing.
+
+e.g.
+
+```json
+"aliases": {
+  "styles": "./{filename}.scss"
+}
+```
+
+will for a file `foo/bar.js` result in
+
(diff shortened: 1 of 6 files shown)


lencioni:
:+1:

PR #79, opened 2015-12-22
Take 'use strict' into account when importing

### LINKED ISSUE #77: Messes up 'use strict';
When adding `'use strict';` to the top of a file to enable strict mode for the file, the `'use strict';` declaration must be at the top of the file. However, ImportJS puts its imports above this declaration, causing it to not work.

trotzig:
The literal expression "use strict" need to be at the top of the file.
Before this commit, we would put imports above it:

  var foo = require('foo');

  'use strict';

This commit will fix that and put imports below:

  'use strict';
  var foo = require('foo');

[Fixes #77]

### DIFF
--- a/lib/import_js/importer.rb
+++ b/lib/import_js/importer.rb
@@ -28,7 +28,9 @@ module ImportJS
 
       old_imports = find_current_imports
       inject_js_module(variable_name, js_module, old_imports[:imports])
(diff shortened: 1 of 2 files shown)


lencioni:
:+1: Thanks for the fast turnaround on these issues!

trotzig:
You know me - always waiting for the next feature request to come in... :)

PR #82, opened 2015-12-24
Update Emacs instructions to use MELPA

kevinkehl:
Allows users to use a package manager to install import-js. Also, fix some escaping.


trotzig:
Looks good. Does this require a new gem being built? I'm assuming it doesn't.

PR #84, opened 2015-12-26
Add plugin for Sublime Text

### LINKED ISSUE #67: Add sublime text plugin
Now that import-js has extracted the core into a standalone gem, it can be integrated more easily with additional editors. We already have emacs support (#61) and we should add sublime text support.

janpaul123:
Which simply always does import_all (which seemed to work well enough on the Brigade codebase), and in ambiguous cases just uses the first match (for now).
    
Uses a command line client that is called from the Sublime Text Python plugin.
    
Also added Command Palette and menu entries, and a readme.

With help from @shreeya for figuring out how to call import-js from Python.

Closes #67.


trotzig:
This is great, thanks for working on it! 

I've been thinking about ways to make the command line tool a little bit more general. By returning json instead of the file contents, we could minimize the amount of data sent back, and start supporting more complex cases (like selecting from a list, passing messages, etc). Here are some examples of what I'm thinking about:

```
⨠ cat Button/index.jsx | import-js Button
```
 would result in
```json
{
  "messages": [
    "Imported 'Foo/bar.js'"
  ],
  "changes": [
    { 
      "type": "add_line",
      "lineNumber": 0,
      "content": "const bar = require('Foo/bar.js');"
    }
  ]
}
```

Or, if there are multiple `bar`s to import, this could be returned:
```json
{
  "askForSelection": [
    "Foo/bar.js",
    "Bar/bar.jsx"
  ]
}
```

In such case, we need to prompt the user for input, then re-run the command with the selection passed in, e.g.

```
⨠ cat Button/index.jsx | import-js Button --select 0
```

trotzig:
I'm happy to keep working on this unless you plan on doing it yourself. I see a few steps going forward:

1. Use the already existing command line editor before merging this PR
2. Let the command-line tool return JSON
3. Add support for importing a word
4. Add support for selection lists

If it's okay with you, could you take on 1? Then I'll work on 2-4, which are more involved. When those are done, It should be possible to let the Emacs plugin use the same command-line tool, and possibly even the VIM plugin.

janpaul123:
I honestly thing we should spend our time wisely. I don't think any of the points 1-4 are more useful than other things we could be doing, like working on Brigade features or a better Controller abstraction or playing in the snow. If you want to spend more time on this, be my guest, but let's do it in follow-up PRs so that people can start using this. :smiley:

janpaul123:
It might be worthwhile to write a blog post about import-js though; we have quite a few integrations now (even if perhaps somewhat incomplete) that we should start spreading the word!

trotzig:
Sounds good. I agree about shipping this and that we have more important things to work on. Though this is a project that I always work on in my spare time, so don't expect me to be too rational about prioritizing... :)

I hope to write the blog post soon!

janpaul123:
:-)

Thanks for merging!

PR #86, opened 2015-12-30
Sublime improvements

trotzig:


trotzig:
@janpaul123 It would be cool if you could try this out. I'm going to publish gem version 0.1.0 next.

PR #93, opened 2016-01-03
Add fix_imports API to Emacs plugin

kevinkehl:
Allow users to use the 'fix_imports' function that automatically adds
unimported variables and removes unused imports. Fix a literal '\n' to
get file contents.

Add instructions on setup. The setup can be a little dicey and is
dependent on your eslint config, which I think is a bit dangerous, so
calling it 'optional'.

Remove references to 'experimental' for Emacs (its not really
experimental anymore =D).


trotzig:
Thanks @kevinkehl!

trotzig:
I'll carve a new gem release tomorrow

PR #94, opened 2016-01-03
Use eslint unix format instead of compact format

lencioni:
Although these formatters produce pretty similar output, it seems that
the unix formatter is made to be machine readable whereas the compact
one may not have been designed with that goal in mind. This should help
this portion of ImportJS be more stable as things change.

There are other machine readable formatters available that I considered,
such as json, but since unix was so simple I decided to go with it.

Thanks to @michaelficarra for the nudge [0].

[0]: https://github.com/eslint/eslint/issues/4845#issuecomment-168501409

### DIFF
--- a/lib/import_js/importer.rb
+++ b/lib/import_js/importer.rb
@@ -105,14 +105,14 @@ module ImportJS
       command = %w[
         eslint
         --stdin
-        --format compact
+        --format unix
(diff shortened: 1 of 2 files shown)


trotzig:
Looks good, thanks @lencioni!

PR #98, opened 2016-01-05
Improve Travis build matrix

lencioni:


lencioni:
Looks like this won't pass until https://github.com/travis-ci/travis-ci/issues/5361 is resolved. In the meantime we could use 2.3.0 instead of 2.3, but I think we may as well just wait.

trotzig:
Sounds good. This isn't likely to rot much even if we have to keep it around. Will you keep an eye out for resolutions to that travis-ci issue?

lencioni:
Sure!

lencioni:
I think this has actually rotted and I'm tired of looking at it, so I'm just going to close this for now. I plan to push a new one once it is working.

PR #100, opened 2016-01-06
Fixed the dropdown insertion + Cancel dropdown insertion added

vini175pa:
As discussed on https://github.com/trotzig/import-js/issues/96 here are my changes on sublime package.

  -  When there were more than 1 one option to import, it was showing a dropdown that was not working due an error. **Fixed** adding the command replace to work asynchronously with `run_command`

- When none option was selected or `esc` was pressed it was importing the last option. **Fixed** adding the conditional `if( i == -1) return`

### DIFF
--- a/import-js.py
+++ b/import-js.py
@@ -19,6 +19,11 @@ def no_executable_error(executable):
     'from the command line.'
   )
 
+class ReplaceCommand(sublime_plugin.TextCommand):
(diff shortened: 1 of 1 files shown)


trotzig:
Great, thanks for contributing! I ended up adding a few commits on top of yours with some style improvements and a bug fix (if multiple selections were presented, it would stop all when you hit escape).

PR #103, opened 2016-01-06
Improve importing defaults and destructured from the same modules

### LINKED ISSUE #83: Importing default and destructuring unnecessarily creates two separate imports when using import declaration keyword
Currently, with `aliases` configuration like the following:

```json
"React": {
  "path": "react",
  "destructure": [
    "PropTypes"
  ]
}
```

importing both `React` and `PropTypes` will result in:

```js
import React from 'react';
import { PropTypes } from 'react';
```

but it would be better if it instead resulted in:

```js
import React, { PropTypes } from 'react';
```

I'm not sure if there is any equivalent syntax for CommonJS imports, but it would be really nice if ImportJS did this when using the `import` declaration keyword. I suppose if you were using CommonJS, it would be better to have one `require` and one assignment after the fact than two `require`s:

```js
const React = require('react');
const { PropTypes } = React;
```

lencioni:
Fixes #83.

### DIFF
--- a/autoload/importjs.vim
+++ b/autoload/importjs.vim
@@ -1,7 +1,6 @@
 function importjs#ImportJSImport()
   ruby $import_js.import
 endfunction
-endfunction
 function importjs#ImportJSGoTo()
   ruby $import_js.goto
 endfunction
--- a/lib/import_js/import_statement.rb
+++ b/lib/import_js/import_statement.rb
@@ -28,7 +28,8 @@ module ImportJS
 
     attr_accessor :assignment
     attr_accessor :original_import_string # a cache of the parsed import string
-    attr_accessor :variables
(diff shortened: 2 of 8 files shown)


lencioni:
I still haven't resolved the issue yet. I think that by using `require` instead of the variable when destructuring, it helps by not creating a separate block the second time, but the sort order changes. I think that once I get the sorting to happen correctly the first time we'll be good to go. Stand by.

lencioni:
Alright, I think this is good to go now.

trotzig:
Agreed. I'll merge when Travis gives her +1.

lencioni:
Travis is green!

trotzig:
wow, you're faster than Travis yourself...

PR #104, opened 2016-01-07
Extract ImportStatement merging to ImportStatement class

lencioni:


lencioni:
Green!

trotzig:
is it worth publishing a new version you think?

lencioni:
Probably. I'm not sure the sorting or deduping will actually matter in most real world situations, but since it is now fixed it wouldn't be a bad idea to get it out there.

trotzig:
I should add:
- vim plugins automatically (depends on setup) get the latest core ruby code
- Package Control (Sublime plugin) looks at tags (semver-looking ones) to decide what to install. Sublime users also need to keep the `import_js` gem in sync. 
- Not sure about the Emacs plugin...

PR #109, opened 2016-01-08
Destructured wrapping

lencioni:


lencioni:
Ah bummer. This fails on Ruby 1.9.3 because I am using keyword arguments. What would you prefer: I replace keyword args with regular args or I remove 1.9.3 from the build matrix?

trotzig:
Well, since I added `slop` a while back, I think we can't use 1.9.3 anyway. At least based on the error message reported at the top of https://github.com/trotzig/import-js/issues/96

You have my blessing.

trotzig:
let me know when you are done

lencioni:
@trotzig done!

trotzig:
I made you a collaborator, so feel free to merge at will!

lencioni:
Sweet, thanks!

PR #114, opened 2016-01-08
Improve some error reporting

lencioni:


lencioni:
I'm not sure under what circumstances eslint will output to stderr, but it might not be a bad idea to bubble stderr up to the editor in every case.

PR #116, opened 2016-01-09
Prevent unnecessarily dirtying the buffer when replacing imports

lencioni:
I noticed that running the fix imports command on a fresh buffer that
didn't actually need it caused the buffer to become dirty even though
nothing changed. This doesn't feel quite right to me, so I decided to
add a bit of code to avoid it.


trotzig:
:+1:

PR #131, opened 2016-01-16
Improve readability of find command in Importer

lencioni:
Instead of using an array of interpolated strings which require escaping
double quotes, we can just use %W[...]. This matches how we build the
eslint command, which I switched to %w[...] in e60c556f.

### DIFF
--- a/lib/import_js/importer.rb
+++ b/lib/import_js/importer.rb
@@ -250,10 +250,10 @@ module ImportJS
         "egrep -i \"(/|^)#{formatted_to_regex(variable_name)}(/index)?(/package)?\.js.*\""
       matched_modules = []
       @config.get('lookup_paths').each do |lookup_path|
-        find_command = [
-          "find #{lookup_path}",
-          "-name \"**.js*\"",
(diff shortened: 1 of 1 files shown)


lib/import_js/importer.rb
```diff
@@ -250,10 +250,10 @@ def find_js_modules(variable_name)
...
       matched_modules = []
       @config.get('lookup_paths').each do |lookup_path|
-        find_command = [
-          "find #{lookup_path}",
-          "-name \"**.js*\"",
-          "-not -path \"./node_modules/*\""
+        find_command = %W[
+          find #{lookup_path}
+          -name "**.js*"
+          -not -path "./node_modules/*"
```
trotzig: In Python, there's a way to pass in an array of tokens forming a command to Popen. I wonder if Ruby has that too? I think each token is then sanitized before being passed down to the command. I'm asking because I think we're somewhat vulnerable to command-injection. What if someone configures a lookup_path that looks like this: `. -name "blah"; rm -rf /;`?

trotzig:
Thanks!

PR #134, opened 2016-01-18
Fix dot lookup paths

lencioni:


lencioni:
I'm going to merge this in. Feel free to leave comments anyway if you like and I can follow up.

trotzig:
Nice catch, and well fixed!

PR #136, opened 2016-01-18
Use FindError instead of ParseError when failing find

lencioni:
As @trotzig pointed out, ParseError isn't ideal here. So I decided to
replace it with a FindError and add a test to verify that it works.

### DIFF
--- a/lib/import_js.rb
+++ b/lib/import_js.rb
@@ -6,6 +6,10 @@ module ImportJS
   class ParseError < StandardError
     # Error thrown when a JS file can't be parsed
   end
+
+  class FindError < StandardError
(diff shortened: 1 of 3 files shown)


trotzig:
Looks like you added some other stuff in the commit too.

lencioni:
What other stuff?

It looks like my new spec failed in travis, but it passes locally. Not really sure why yet, but I would guess that `find` behaves differently there than it is on my laptop.

lencioni:
I added a check to fail if lookup path is an empty string, so specs should pass now.

trotzig:
I was confused. Looks good!

PR #146, opened 2016-01-23
Auto-import React if implicit dependency exists

### LINKED ISSUE #113: Auto-fix missing invisible React import when using JSX
If you are using JSX, you need to import React even though the variable doesn't appear to be used. [eslint-plugin-react has a rule that detects this for you](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/react-in-jsx-scope.md), which we could piggyback off of.

This should probably be enabled (or disabled?) as an option, since some people might be defining `React` as a global variable.

trotzig:
By allowing import-js to know about eslint output from a specific plugin
(react-in-jsx-scope [1]). This isn't great, but it solves an important
usecase when using jsx, namely that `React` must be in scope when trying
to do stuff like

  import 'Button' from 'components/button';
  const button = <Button/>;

I added a section to the Readme to explain how this feature works. While
doing that, I improved the section about eslint a little bit.

Fixes #113

[1]: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/react-in-jsx-scope.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,13 @@ automatically fix all imports. By hiting `<leader>i` (Vim), `(M-x)
 import-js-fix` (Emacs), or choose `ImportJS: fix all imports` (Sublime), all
 your undefined variables will be resolved, and all your unused imports will be
 removed. By default, import-js expects a global `eslint` command to be
-available.
+available, but you can change that with the [`eslint_executable` configuration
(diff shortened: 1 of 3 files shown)


lencioni:
Just one clarification in the readme, otherwise :+1:

PR #147, opened 2016-01-23
Import dev dependencies

trotzig:


lencioni:
Question about local configurations, but otherwise :+1:

PR #149, opened 2016-01-24
Add .rubocop.yml, fix many rubocop errors

lencioni:
We want to improve the code quality an consistency of this codebase, so
we are deciding to use rubocop. This commit fixes most of the things
that it was complaining about and adds a little conservative
configuration to get us started. There are still a number of errors that
it is complaining about that we still need to deal with, either by
fixing or adjusting configuration, and there might be other rubocop
rules that are disabled by default that we will want to enable, but this
seems like a good first pass.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,17 @@
+AllCops:
+  DisplayCopNames: true
+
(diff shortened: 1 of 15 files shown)


trotzig:
Thanks!

PR #157, opened 2016-01-25
Add 0.3.0 to messages.json

lencioni:
For the benefit of Sublime Text users.


trotzig:
Thanks! I forgot about doing that!

PR #158, opened 2016-01-25
Break out `destructures` from `aliases`

### LINKED ISSUE #135: Split out destructuring configuration from aliases
Since adding destructuring to aliases configuration, the name `aliases` is overly specific. I think we should rename it to something that better captures how the configuration is actually used. `modules` comes to mind, but I'm open to other suggestions. And, when we make this change, I think we should make `path` optional. This would clear up some redundant configuration, like:

```json
"aliases": {
  "React": {
    "path": "react",
    "destructure": [
      "Children",
      "Component",
      "PropTypes"
    ]
  }
}
```

would then become:

```json
"modules": {
  "React": {
    "destructure": [
      "Children",
      "Component",
      "PropTypes"
    ]
  }
}
```

trotzig:
Some of the destructures you use will need an alias, but not all.
Before this change, in order to add a destructure, you also needed to
add an alias. To make this process easier to understand and more
powerful, we've decided to break out destructures into its own thing.
Easier because you'll no longer see aliases being objects mixed with
aliases being strings. More powerful because it is now more obvious that
you can use destructures for non-package-dependencies (this was possible
before, but they were awkwardly placed under the `aliases`
configuration).

Fixes #135

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,29 +159,6 @@ those, you can add them to the `aliases` configuration.
 }
 ```
 
-If you have a library that expose a single object that has a bunch of objects
-on it that you want to use, you can list those in a `destructure` array inside
-the alias (which then has to be turned into an object):
-
-```json
-"aliases": {
-  "$": "third-party-libs/jquery",
-  "_": {
-    "path": "third-party-libs/underscore",
-    "destructure": ["memoize", "debounce"]
-  }
-}
-```
-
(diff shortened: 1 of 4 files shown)


lencioni:
Thanks for pounding this out on your skiing trip! :ski: :ski: :ski:

PR #162, opened 2016-01-26
Rename `destructures` to `named_exports`

### LINKED ISSUE #161: Rename `destructures` to `named_exports` or `named_imports`
ES2015 modules can have default exports and named exports (along with default imports and named imports). In keeping with our trend of ES2015-first (e.g. defaulting `declaration_keyword` to `'import'`), I'm wondering if we should rename the `destructures` configuration key to something like `named_exports` or `named_imports`. What do you think?

lencioni:
ES2015 modules can have default exports and named exports (along with
default imports and named imports). In keeping with our trend of
ES2015-first (e.g. defaulting declaration_keyword to 'import'), I'm
renaming the `destructures` configuration key to `named_exports`, and
adjusting the names we use in the code to match.

This is a little bit of a blurry line, since the rules are actually
different between ES2015 named exports and CommonJS destructuring, and
if you are using ES2015 `import` on a CommonJS module that has things
defined in the `named_exports` configuration, that doesn't actually make
them named exports and according to the spec shouldn't actually work.
You might have the inverse problem if not using `import` and requiring
an ES2015 module. In practice, this likely depends on what tool you use
to build your modules and the compatibility layer they've decided to
cook up for ESM/CommonJS interoperability.

Since I think this issue exists for either name we choose, I think we
should go with the one that is most consistent with our approach, which
is `named_exports`. Alternatively, we could use both `named_exports` and
`destructures`, but I'm not sure that's really worth it.

Fixes #161

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,15 +177,15 @@ import styles from './bar.scss';
 ```
 
 
-### `destructures`
+### `named_exports`
 
 If you have a ES2015 module that exports multiple things (named exports), or a
 CommonJS module that exports an object with properties on it that you want to
-destructure when importing, you can add those to a `destructures` configuration
+destructure when importing, you can add those to a `named_exports` configuration
 option.
 
 ```json
-"destructures": {
+"named_exports": {
   "underscore": [
     "omit",
     "debounce"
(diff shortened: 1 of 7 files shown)


trotzig:
:+1: 

Sounds good. Thanks for sharing that bit of knowledge. I wasn't aware of all the differences between commonjs and es2015 exports.

PR #164, opened 2016-01-26
Check version used against `minimum_version` config

### LINKED ISSUE #153: Help people understand when they need to upgrade import-js
If the format of the configuration file changes in a way that requires people to upgrade their editor plugins, it would be great if there was a built-in way to notify people that they need to do this.

I'm thinking that the simplest solution would be to add a line to the configuration file that says the minimum version of the plugin that is required. Then, when import-js runs, it can check its version against this value and if it isn't new enough, it can tell the person that they need to update their plugin.

Unfortunately, this doesn't work very well for the other direction, where the plugin is too new for the configuration version, but that seems less likely to cause major problems. If we wanted to solve for this, we could probably do something like semver ranges. I think this would probably work okay, but I worry that it would be unnecessarily confusing and cause too much churn in the configuration file.

Ideally, we'd figure out exactly how we want to solve this as soon as possible so it can start benefiting people as soon as the next release.

trotzig:
If the format of the configuration file changes in a way that requires
people to upgrade their editor plugins, we will now help people
understand that they need to upgrade by showing an error message.

The current version of import-js will be checked against a
`minimum_version` configuration. If the version requirement isn't
fulfilled, an error is thrown. I'm using the built-in `Gem::Dependency`
class to check versions. Right now I'm prepending `>=` to the min
version before checking. If we find that this is too simple, we can
choose to expand the `minimum_version` to a `match_version` (or
something) and allow people to use a gem-dependency-style matching
string instead, e.g. `~> 1.2.3`, `> 2.0`.

Fixes #153

### DIFF
--- a/lib/import_js.rb
+++ b/lib/import_js.rb
@@ -10,6 +10,10 @@ module ImportJS
   class FindError < StandardError
     # Error thrown when the find command fails
   end
+
+  class ClientTooOldError < StandardError
+    # Error thrwon when the client is too old to handle the config
+  end
 end
 
 require_relative 'import_js/command_line_editor'
--- a/lib/import_js/configuration.rb
+++ b/lib/import_js/configuration.rb
@@ -15,6 +15,7 @@ module ImportJS
     'import_dev_dependencies' => false,
     'import_function' => 'require',
     'lookup_paths' => ['.'],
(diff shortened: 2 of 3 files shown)


lencioni:
:+1:

PR #180, opened 2016-01-31
Remove `let` as valid declaration_keyword choice

### LINKED ISSUE #163: Drop support for `let` as `declaration_keyword`
I can't think of a scenario where someone would want to actually use `let` as `declaration_keyword`. Although supporting it is very easy for us in code, I think that having the option available creates unnecessary confusion for people getting this set up.

lencioni:
I can't think of a scenario where someone would want to actually use let
as declaration_keyword. Although supporting it is very easy for us in
code, I think that having the option available creates unnecessary
confusion for people getting this set up.

In practice, we still need to support matching `let` in our regexes
because people can type whatever they want and we don't want to fall
over if they decide to use `let` for some reason. So, this mostly
removes its usage from documentation and comments.

Fixes #163

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,7 +206,8 @@ import { memoize } from 'underscore';
 memoize(() => { foo() });
 ```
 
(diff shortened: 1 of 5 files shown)


trotzig:
Thanks!

PR #185, opened 2016-01-31
Add dispatch settings to .projections.json

lencioni:
This tells plugins like vim-dispatch what to do by default when dispatch
is called without arguments, making it easier to run tests on the
current file.


trotzig:
:+1: I should start using this.

PR #188, opened 2016-02-02
Default Vim's max_line_length to 80

### LINKED ISSUE #187: imports were reordered, react was added on two lines
Trying this out on vim and I had

```js
import Traps from 'traps';
import BoosterSeats from 'booster-seat-data';
import cx from 'classnames';
import CarMixin from '../../../components/CarMixin';

const PhotoCarousel = React.createClass({ ... });
```

I put my cursor under `React`, `leader-j`, and this happened to the imports.

```js
import BoosterSeats from 'booster-seat-data';
import CarMixin from '../../../components/CarMixin';
import React from
  'react';
import Traps from 'traps';
import cx from 'classnames';
```

* React is on two lines instead of one?
* The imports were re-ordered. While I normally don't have an issue with this convention, this can lead to noisy diffs. Also, it's a bit opinionated.

lencioni:
By default, Vim will run with a textwidth of 0. With this setting,
everything ends up getting wrapped, which is not very good. To make this
better, I am defaulting the max_line_length to 80 if we encounter a 0.

In the long run, I think we should move this back into configuration,
since teams use a variety of editors which may not be configured the
same way.

Fixes #187

### DIFF
--- a/lib/import_js/vim_editor.rb
+++ b/lib/import_js/vim_editor.rb
@@ -113,7 +113,9 @@ module ImportJS
     # Get the preferred max length of a line
     # @return [Number?]
     def max_line_length
-      get_number('&textwidth')
+      length = get_number('&textwidth')
+      return length unless length == 0
+      80
     end
 
     # @return [String] shiftwidth number of spaces if expandtab is not set,
--- a/spec/import_js/importer_spec.rb
+++ b/spec/import_js/importer_spec.rb
@@ -1111,6 +1111,46 @@ foo
           EOS
         end
       end
+
+      context "when Vim's textwidth is 0" do
+        # This is the default, unconfigured behavior for Vim
+        before(:each) do
+          allow_any_instance_of(ImportJS::VIMEditor)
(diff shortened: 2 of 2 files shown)


lib/import_js/vim_editor.rb
```diff
@@ -113,7 +113,9 @@ def ask_for_selection(word, alternatives)
     # Get the preferred max length of a line
     # @return [Number?]
     def max_line_length
-      get_number('&textwidth')
+      length = get_number('&textwidth')
+      return length unless length == 0
```
trotzig: nbd, but I think this could have been a one-liner: `get_number('&textwidth') || 80`

lencioni: Nope, 0 is truthy in ruby.

```
[1] pry(main)> 0 || 80
=> 0
```

trotzig: Well, then just wrap it in an external call to some node.js script. ;)

Issue #173, opened 2016-01-30
Add ability to run force fix imports on multiple files from the command line


lencioni:
If you are working in a codebase that decides to switch from `const` to `import` syntax, it would be nice if there was a built-in way to run a command that would do all of this work for you.

lencioni:
And maybe this shouldn't add or remove imports, but just update the current imports to match the style specified in the config. This should include things like declaration keyword, custom require function, excluding extensions, combining named exports, grouping (once that's a feature).

trotzig:
A tricky thing here is that in order to resolve the configuration for an import, we sort of need it's full file path. The import statement itself isn't enough in some cases:

```js
const Foo = require('foo');
```

If we have local configuration for `app/**`, how do we know if `Foo` belongs to that local configuration or not? I guess we can do what `goto` does - try to resolve the import and out of the resolved `JSModule`s for `Foo`, check which ones end up in `'foo'` as the import path. Then use the file path from the `JSModule` to check config.

trotzig:
As for batch-updating, I think we should just allow people to do something like this:

```bash
find . -name "*.js*" -exec import-js --filename {} --overwrite \;
```

Then we can just focus on making the command-line tool work well for one file.

lencioni:
Yeah, that's fine. We can always add a command that makes that easier later if we need it.

trotzig:
I've been thinking about this a little bit, and I think this is what we should do to begin with:

Introduce a new command alongside `fix_imports`, called `rewrite_imports`. We parse the file for current imports, then try to resolve each variable from those imports again, one by one. We'll use the same technique that `goto` uses in order to map ambiguous imports against the existing import statements. 

This technique should be pretty robust. The downside is that it will probably be slow, at least if we don't do any optimization.

lencioni:
Slow is better than non-existent. That sounds like a fine starting point to me.

PR #191, opened 2016-02-03
Allow command-line tool to overwrite file contents

trotzig:
We want to support batch-updating multiple files via the `import-js`
command. An important part of that is to overwrite the file with the new
content. This commit takes care of that, and at the same time adds
support for passing in only a --filename and no content via stdin.

This takes us one step closer to supporting the usecase described in
[#173].

### DIFF
--- a/README.md
+++ b/README.md
@@ -407,19 +407,20 @@ editor](SUBLIME.md) uses.
 
 ```bash
 ⨠ import-js --help
-usage: import-js [options]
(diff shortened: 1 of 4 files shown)


trotzig:
This isn't really urgent, so don't feel stressed about looking at it @lencioni.

trotzig:
@lencioni This is now ready for a second review (if you have time). :)

lencioni:
:+1:

PR #193, opened 2016-02-06
Add `rewrite_imports` command

### LINKED ISSUE #173: Add ability to run force fix imports on multiple files from the command line

trotzig:
Invoked by using the `--rewrite` flag in the CLI tool, this command will
rewrite all existing import statements. This is useful if you want to
batch-update many files with a change to `declaration_keyword`,
`use_relative_paths`, `import_function` etc.

I'm not documenting this feature too much at this point. There are
a few tricky edge-cases that I still haven't figured out how to deal
with:

 - Resolving ambiguous imports - we currently just print some info to
   the JSON object output to stderr.
 - Switching to relative imports doesn't currently work - we need to
   figure out a better way to get rid of duplicates in the resulting
   import statements than just using `uniq`.
 - Current imports that we can't map to a js module - this can happen if
   you have manually entered import statements. Right now we just keep
   them as-is.

[Fixes #173]

### DIFF
--- a/bin/import-js
+++ b/bin/import-js
@@ -18,6 +18,8 @@ opts = Slop.parse do |o|
          'behavior is to print to stdout). This only applies if you are ' \
          'passing in a file (<path-to-file>) as the first positional argument.'
   o.string '--filename', '(deprecated) Alias for --stdin-file-path'
+  o.bool '--rewrite',
+         'Rewrite all current imports instead of importing anything.'
 
   o.on '-v', '--version', 'Prints the current version' do
     puts ImportJS::VERSION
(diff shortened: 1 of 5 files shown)


trotzig:
@lencioni specs will fail right now, though feel free to give this a test drive! I had a go at the Brigade codebase and it works pretty well.

lencioni:
Looking good! Thanks for working on this. :)

trotzig:
I'm pretty sure there are a few subtle bugs/inconveniences in this code. But until this tool has come to use, it's hard to tell what situations we will run into. I'm going to suggest we fix problems as they appear when we start using this instead of trying to make this ultra-robust from the start.

lencioni:
Makes sense! I'll try to take it for a spin tomorrow and see how it goes. I have a project that has a lot of legacy code, so it is likely to turn up a few things. :)

PR #200, opened 2016-02-27
Group imports into blocks based on a heuristic

### LINKED ISSUE #138: Group imports into separate blocks
It might be nice, perhaps as an option, to separate packages and local modules in the import block. What I mean by this is instead of having one import block that is all sorted alphabetically, import-js would manage two import blocks that are individually sorted alphabetically. The first would include any package dependencies, and the second would include any local dependencies.

Example:

```js
import _ from 'underscore';
import React from 'react';

import MyComponent from './MyComponent';
import SomeOtherThing from '../../SomeOtherThing';

// rest of file would go here
```

In files that have a lot of dependencies, this might help keep things a bit more organized and readable.

lencioni:
It might be nice, perhaps as an option, to separate packages and local
modules in the import block. What I mean by this is instead of having
one import block that is all sorted alphabetically, import-js would
manage multiple import blocks that are individually sorted alphabetically.
The first would include any package dependencies, and the second would
include any local dependencies.

Example:

```js
import _ from 'underscore';
import React from 'react';

import MyComponent from './MyComponent';
import SomeOtherThing from '../../SomeOtherThing';

// rest of file would go here
```

In files that have a lot of dependencies, this should help keep things a
bit more organized and readable.

We considered making each group configurable, but ran into issues with
being able to map import statements back to actual paths
deterministically, so we decided to use a heuristic based on the style
of the imports and the paths. In most situations this will produce 2
groups (one for packages and one for non-packages), but in theory there
could be up to 12 groups.

We discussed making this configurable, but I wanted to get feedback
about my current implementation before doing that work.

Fixes #138


spec/import_js/importer_spec.rb
```diff
@@ -1331,8 +1332,8 @@ def self.current_selection=(index)
             it 'adds the destructuring on a new line' do
               expect(subject).to eq(<<-EOS.strip)
 var _ = require('underscore');
-var foo = require('foo');
 var { memoize } = require('underscore');
+var foo = require('foo');
```
lencioni: A nice side-effect of this is now we are sorting based on `ImportStatement`s instead of the strings that they generate, which allows destructured requires to be next to the default requires.

trotzig:
Thanks for all the work on this! I'm going to do some manual testing and check with the team before getting back to you about merging.

trotzig:
I've tried this out on the Brigade codebase, and it looks pretty decent! What's your opinion on making it configurable? I could go both ways, though I could see this being irritating if you don't want it and can't turn it off. Maybe `group_imports`, where the default is `true`?

lencioni:
Awesome! I think we should only add the config if it is actually needed. If
you need it at Brigade or if someone else needs it then let's add it.
Otherwise let's keep it slim.

trotzig:
@lencioni FYI - I pushed a bugfix for an issue I found when running the CLI tool.

trotzig:
I have a preliminary +1 from the team to group imports. I'll follow up in the coming days when I have a few extra cycles. I'll probably merge this followed by releasing 0.5.0. It's probably a good idea to update your `.importjs.json` at that time with a `minimum_version` to force everyone to upgrade.

trotzig:
I ended up pushing another bugfix that I found when rewriting the Brigade codebase. Not introduced in this branch though, but we might as well put it here.

PR #205, opened 2016-03-02
Fix unused variables sharing names of imports

### LINKED ISSUE #203: Unused variables that share the names of imports confuse fix imports
If you have code that looks something like the following:

```js
import uuid from 'uuid';

function bar() {
  return uuid.v4();
}

export default function foo() {
  const things = {
    uuid: bar(),
    henric: 'is cool',
  };

  const { uuid, henric } = things;
  return henric;
}
```

Running fix imports will toggle between adding and removing the `uuid` import. This is because the error that eslint reports relates to the `uuid` defined by `const { uuid, henric } = things;` line. However, after it is removed, we get a second error saying that it is used but never defined (by `return uuid.v4()`), so it gets added back again.

lencioni:
If you have code that looks something like the following:

```js
import uuid from 'uuid';

function bar() {
  return uuid.v4();
}

export default function foo() {
  const things = {
    uuid: bar(),
    henric: 'is cool',
  };

  const { uuid, henric } = things;
  return henric;
}
```

Running fix imports will toggle between adding and removing the `uuid`
import. This is because the error that eslint reports relates to the
`uuid` defined by `const { uuid, henric } = things;` line. However,
after it is removed, we get a second error saying that it is used but
never defined (by `return uuid.v4()`), so it gets added back again.

To fix this, I added some logic that verifies that the eslint error for
unused variables happens within the imports block.

While I was working on the fix_imports method, I found a couple of
opportunities to return early to avoid doing unnecessary work.

Fixes #203

### DIFF
--- a/lib/import_js/importer.rb
+++ b/lib/import_js/importer.rb
@@ -99,8 +99,7 @@ module ImportJS
 
       # Filter out unused variables that do not appear within the imports block.
       unused_variables.select! do |_, line_numbers|
-        any_line_numbers_within_imports_range?(
-          line_numbers, old_imports[:range])
+        any_numbers_within_range?(line_numbers, old_imports[:range])
       end
 
       new_imports = old_imports[:imports].clone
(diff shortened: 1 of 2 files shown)


lib/import_js/importer.rb
```diff
@@ -130,6 +145,16 @@ def message(str)
...
 
+    # @param numbers [Set]
+    # @param range [Range]
+    # @return [Boolean]
+    def any_numbers_within_range?(numbers, range)
+      numbers.each do |number|
+        return true if range.include?(number)
+      end
+      false
+    end
```
trotzig: This could have been inlined with the other stuff, I found the indirection made it harder to follow. No big deal though.

trotzig:
Thanks!

lib/import_js/importer.rb
lencioni: I originally had it that way, but I think I was running into weirdness with not being able to return from the block, since that will return from the function. I suppose I could have written this as a `find` instead of a `each`, which would have helped. Feel free to rework.

PR #206, opened 2016-03-03
Vim: Map commands only in JavaScript files

lencioni:
Vim has decent filetype detection built into it, which enables files in
the special ftplugin directory to be applied to files of specific types.
I think it makes the most sense for us to put our mappings in a
ftplugin/javascript.vim file so that they are only set up when in
JavaScript files.

In making this change, I also had to add the <buffer> argument to the
mappings, which makes the mapping effective in the current buffer only.
This will prevent the mapping from leaking out into other types of files
after visiting a JavaScript file.

### DIFF
--- a/ftplugin/javascript.vim
+++ b/ftplugin/javascript.vim
@@ -1,0 +1,11 @@
+if !hasmapto(':ImportJSImport<CR>') && maparg('<Leader>j', 'n') == ''
+  silent! nnoremap <buffer> <unique> <silent> <Leader>j :ImportJSImport<CR>
+endif
+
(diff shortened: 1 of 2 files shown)


lencioni:
@wincent I'm curious about what you think about this change. Could you take a look when you have a moment?

wincent:
Makes sense to keep the impact low by restricting it to JS files only. Might also want to use <Plug> mappings moving forward, as those are considered a little more useful: http://whileimautomaton.net/2008/09/27022735

trotzig:
I added a new issue to cover the transition to <Plug> mappings.

trotzig:
Thanks @lencioni!

PR #211, opened 2016-03-03
Put imports for package-local modules in right group

### LINKED ISSUE #208: Peek in node modules directories when grouping
If you depend on a package `foo` but end up importing a file that exists within the package, such as `foo/some/other/thing`, you might write:

```js
import thing from 'foo/some/other/thing';
```

In this case, we not currently resolving this as a package dependency when grouping, but we should.

trotzig:
Me and @lencioni have both come across examples where we import a
package-local module (through an alias). These used to end up in the
wrong import group, because of how we were matching them against package
dependencies. Here's an example from the Brigade codebase:

const Nprogress = require('nprogress/nprogress');

By stripping out everything after and including the first slash before
matching against package dependencies, they are back in the right group
again.

[Fixes #208]

### DIFF
--- a/lib/import_js/import_statements.rb
+++ b/lib/import_js/import_statements.rb
@@ -166,7 +166,11 @@ module ImportJS
     # @param package_dependencies [Array<String>]
     # @return [String] 'package, 'non-relative', 'relative'
     def import_statement_path_type(import_statement, package_dependencies)
-      path = import_statement.path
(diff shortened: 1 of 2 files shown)


lencioni:
:+1:

lencioni:
Geepers, Travis sure is taking its time today.

trotzig:
Galushi indeed.

PR #212, opened 2016-03-03
Use \S instead of [^\s]

trotzig:
As pointed out by @lencioni.

https://github.com/trotzig/import-js/commit/bd696a0da66af65dd9f05365eaf8297db17ba61d#commitcomment-16481843


lencioni:
:+1:

PR #213, opened 2016-03-03
Clean out whitespace when removing the last import

### LINKED ISSUE #204: Removing final import with fix imports leaves leading whitespace
If you have a file like:

```js
import Foo from 'foo';

// rest of code here
```

And `Foo` ends up being removed from the imports, leaving no more imports, you will currently (at least on the groupings branch) end up with

```

// rest of code here
```

I would expect this to not leave that leading newline instead:

```js
// rest of code here
```

trotzig:
Normally, we want a blank line after the block of imports. When removing
the last import however, we would leave an unwanted blank line behind.

[Fixes #204]

### DIFF
--- a/lib/import_js/importer.rb
+++ b/lib/import_js/importer.rb
@@ -247,9 +247,9 @@ module ImportJS
 
       if import_strings.empty? &&
          @editor.read_line(old_imports_range.first).empty?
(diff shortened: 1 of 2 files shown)


lencioni:
:+1:

PR #216, opened 2016-03-04
Allow grouping to be disabled

trotzig:
This commit adds a new `group_imports` configuration option. By setting
this to false, you make all imports end up in one single group, sorted
alphabetically.

After using the grouping for a while on the Brigade codebase, I've
noticed a few issues:

 - import-js bugs are harder to spot (because we have newlines in
   imports, it's hard to spot where parsing of imports stopped)
 - some imports are hard to know what group to put in. e.g. aliases that
   resolve to something inside a package dependency, but doesn't clearly
   say so.
 - merge mistakes happen because people don't understand why the
   whitespace is there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,8 +98,8 @@ the cursor on a variable and hit `<leader>g` (Vim), `(M-x) import-js-goto`
 
 - Only files ending in .js\* are considered when importing
 - As part of resolving imports, all imports will be sorted and placed into
-  groups (package dependencies first, then one or more groups with internal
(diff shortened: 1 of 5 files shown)


lencioni:
I think it might be worth considering this being disabled by default. What do you think?

trotzig:
Yeah, it's only that we changed the default in 0.5.0. Now we would be changing it back again in 0.6.0. Maybe it's okay though.

Issue #220, opened 2016-03-05
Add instructions of how to configure the Sublime plugin to SUBLIME.md


trotzig:
As discovered by @ljharb in https://github.com/trotzig/import-js/issues/219, there's no information how to configure the Sublime plugin in the Sublime docs. We need to inform people how to override the path to the executable, something like:

> If you get an error message saying something like "Can't find import-js executable", you may need to specify a path to the `import-js` executable in configuration. Create or edit /Packages/User/ImportJS.sublime-settings with something that looks like this:
```json
{
  "executable": "/Users/henrictrotzig/.rbenv/shims/import-js"
}
```
Please note that you can't use ~ to refer to the home directory, you need to specify the full path. 
To figure out where your `import-js` executable is located, you can run `which import-js`.

lencioni:
It would be nice to have sections specifically detailing what to do if using rbenv and what to do if using rvm.

Issue #219, opened 2016-03-05
Issues with sublime 3 and rvm?


ljharb:
I tried to install the plugin in Sublime 3, but despite installing the import_js gem in my system ruby, and in all 3 of my rvm-installed rubies, I still get the message that it couldn't find the executable.

The popup suggests adding the path in my user settings - but there's no place under "Package Settings" for import-js, so I don't know where to add it, nor do I know the syntax/structure to add.

trotzig:
To configure the path to the executable, you can follow these instructions (added in the [release notes for 0.2.1](https://github.com/trotzig/import-js/releases/tag/0.2.1)):

> Breaking change: The default Sublime setting for executable has changed to just import-js, from previously being ~/.rbenv/shims/import-js. If you were relying on the rbenv path, you need to override the executable setting with a full path to the import-js executable. Create or edit <path-to-sublime>/Packages/User/ImportJS.sublime-settings with something that looks like this:
```json
{
  "executable": "/Users/henrictrotzig/.rbenv/shims/import-js"
}
```
Please note that you can't use ~ to refer to the home directory, you need to specify the full path.

To figure out where rvm put the `import-js` executable, you can run `which import-js`.

Sorry for not having these in the Sublime docs. That's a mistake, and I'm going to follow up with a fix.

lencioni:
There is also a popup that the sublime plugin generates that could probably be improved as well.

lencioni:
I've improved some documentation, added a menu, and improved an error message, which I think will all help with this. Thanks for reporting it!

ljharb:
thanks!

lencioni:
@ljharb 0.6.0 is out now with these changes. Will you try it out and let us know if you have any more issues getting up and running?

trotzig:
Package Control isn't great at finding new releases directly, so you might have to wait a while (assuming you use Package Control).

https://packagecontrol.io/packages/ImportJS

Still on 0.5.1.

ljharb:
I am using Package Control - i'll let you know after it updates.

trotzig:
Thanks!

lencioni:
Looks like it has been updated!

ljharb:
Thanks! One nit: i think the default user settings should be a copy of the default settings the first time they're opened (currently it's an empty file).

So - I updated, and set my executable to the system ruby's import-js path (`/usr/local/bin/import-js`), and now i get a popup with:
```
Error when executing import-js:

[...]

PR #222, opened 2016-03-05
Add menu items to Sublime to open settings files

lencioni:
This is a common convention for Sublime Text plugins and will make it
easier to modify the settings that may be required to get ImportJS
working with Sublime Text.

For some reason if I put the preferences child above the tools child,
the tools menu item no longer shows up, but in this order things are
just fine. I wasn't able to determine why, so I'm just going to roll
with it.

Addresses #220 and #219

### DIFF
--- a/Main.sublime-menu
+++ b/Main.sublime-menu
@@ -8,5 +8,38 @@
         "command": "import_js"
       }
     ]
+  },
+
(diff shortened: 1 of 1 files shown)


ljharb:
:+1:

PR #224, opened 2016-03-05
Make it easier to import Node.js core modules

### LINKED ISSUE #201: Add support for importing built in Node modules
For instance, if you are writing code targeting Node.js, you might want to use a module that is part of the standard library, such as [`Path`](https://nodejs.org/dist/latest-v4.x/docs/api/path.html) or [`Net`](https://nodejs.org/dist/latest-v4.x/docs/api/net.html). When using something like this, you need to use `require('path')` first. It would be nice if import-js could be configured to resolve these for you.

I think it would probably make sense for these to be in their own group, sorted at the very top (above packages).

trotzig:
We want people that are in a Node environment to be able to import core
modules (like fs, path, readline) quickly. To do that, we're borrowing
ideas from eslint, where you can configure what environments a project
is using [1]. I decided to use `environments` as the config key instead
of `env`, and use an array of environments instead of a hash (env =>
boolean). That made the config option more consistent with how other
keys are named and used.

Imports for core modules will end up at the top of the imports, in its
own group (above) package dependencies.

Fixes #201

[1]: http://eslint.org/docs/user-guide/configuring#specifying-environments

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,17 @@ will for a file `foo/bar.js` result in
 import styles from './bar.scss';
 ```
 
+### `environments`
+
+This list of environments control what core modules are available when
+importing. The only supported value right now is `["node"]`, and if you use it
+you automatically make [all the core modules for
+Node](https://nodejs.org/api/modules.html#modules_core_modules) available for
+import-js.
+
+```json
+"environments": ["node"]
+```
 
 ### `named_exports`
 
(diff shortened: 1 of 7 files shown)


lencioni:
Just to double check, you will be able to turn this on for a subset of a project right?

trotzig:
Yes, based on the file you are editing (`applies_to`). Not based on the file you are importing (`applies_from`).

Issue #186, opened 2016-01-31
Use relative paths by default


lencioni:
Similar to using `import` as the default declaration keyword (#118), I think we should consider using relative paths by default since that is the standard way of doing things. Using non-relative paths requires a tool like webpack to be configured in specific ways and should also probably require additional configuration for Import-JS. What do you think?

trotzig:
Sounds good!

PR #231, opened 2016-03-15
Add Atom plugin

### LINKED ISSUE #101: Add Atom plugin
Now that import-js has extracted the core into a standalone gem, it can be integrated more easily with additional editors. We already have emacs and Sublime Text support and we should add Atom support.

lencioni:
This commit adds a plugin for the Atom editor. There are still some
improvements that can/should be made here, which I believe I mentioned
in TODO comments, but this version fully works.

I considered moving the eslint dependency up a level, but I didn't want
to step on @trotzig's feet while he is porting things to JS.

We should also add tests for this stuff, since it is entirely untested.

https://atom.io/docs/v1.5.4/hacking-atom-writing-specs

We will also need to publish the Atom npm package.

Fixes #101


trotzig:
So exciting! I'll have a look at this soon. Is there a package manager for Atom (similar to Package Control for Sublime)?

CONTRIBUTING.md
```diff
@@ -61,25 +61,47 @@ Make a symlink inside your Sublime packages folder to the local copy of
...
+If you just need to make some small changes to the gem for debugging purposes,
+you can open the installed gem:
+
+```sh
+gem open import_js
+```
+
+Note that any changes made here will only affect the installed gem and not the
+import-js codebase.
+
```
trotzig: Neat, thanks!

atom/.eslintrc.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  extends: 'airbnb/base',
```
trotzig: What goodies do we get from here?

atom/keymaps/import-js.cson
```diff
@@ -0,0 +1,17 @@
...
+# For more detailed documentation see
+# https://atom.io/docs/latest/behind-atom-keymaps-in-depth
+'atom-workspace':
+  'cmd-shift-j': 'import-js:import'
+  'cmd-shift-k': 'import-js:goto'
+  'cmd-shift-i': 'import-js:fix-imports'
+
+# TODO(Joe): I have no idea if these keybindings are any good and didn't give
+# them much thought at all, so we should definitely improve this before
+# shipping.
```
trotzig: Want to improve before shipping? Or just ship? :)

atom/lib/AskForSelectionView.js
```diff
@@ -0,0 +1,60 @@
...
+
+  viewForItem(item) {
+    return `<li>${item}</li>`;
+  }
+
+  confirmed(item) {
+    // The ImportJS CLI tool expects an index from the alternatives list, so we
+    // need to find the index.
+    // TODO: should we change the API so the CLI tool takes a path instead of an
+    // index?
```
trotzig: Yeah, I thought about that when rewriting the importer spec as well. I think it's fine with indexes for now (it also makes the command easier to construct, no need for escaping).

atom/lib/Deferred.js
```diff
@@ -0,0 +1,25 @@
...
+    });
+
+    this.then = function promiseThen() {
+      return this.promise.then.apply(this.promise, arguments);
+    }.bind(this);
+
+    this.catch = function promiseCatch() {
+      this.promise.catch.apply(this.promise, arguments);
+    }.bind(this);
+  }
```
trotzig: I'm surprised that there's no built-in deferred..

atom/lib/getEnv.js
```diff
@@ -0,0 +1,11 @@
...
+
+const consistentEnv = require('consistent-env');
+
+const envPromise = consistentEnv.async()
+  .catch(error => {
+    atom.notifications.addError(error.message);
+    throw error;
+  });
+
+module.exports = () => envPromise;
```
trotzig: Care to explain what this does?

atom/lib/import-js.js
```diff
@@ -0,0 +1,302 @@
+'use strict';
+
+const AskForSelectionView = require('./AskForSelectionView');
```
trotzig: Did you use import-js for these?

atom/lib/import-js.js
```diff
@@ -0,0 +1,302 @@
...
+const getEnv = require('./getEnv');
+
+const BufferedProcess = require('atom').BufferedProcess;
+const CompositeDisposable = require('atom').CompositeDisposable;
```
trotzig: Could these two be deconstructed? 

```js
const { BufferedProcess, CompositeDisposable } = require('atom');
```

atom/lib/import-js.js
```diff
@@ -0,0 +1,302 @@
...
+      // No selections were chosen.
+      return undefined;
+    }
+
+    // Foo:0,Bar:1
+    const selectionsString = selections
+      .map((selection) => `${selection.word}:${selection.alternativeIndex}`)
+      .join(',');
+
+    /* eslint-disable no-use-before-define */
```
trotzig: Use `// eslint-disable-line no-use-before-define` ?

atom/lib/import-js.js
```diff
@@ -0,0 +1,302 @@
...
+ */
+function importWord(selections) {
+  const args = [];
+
+  if (selections) {
+    // We have selections to import.
+    args.push('--selections', selections);
+  } else {
+    // There are no selections to import, so let's figure out what the current
+    // word is.
```
trotzig: I think you should use `--word` when there are selections as well. This is because `import-js` needs to redo the same command again when applying selections. Look at the Sublime plugin for this. 

I think you'll see a bug here if you try to import a single word that ends up asking for selection, then picking a selection.

atom/lib/import-js.js
```diff
@@ -0,0 +1,302 @@
...
+    return;
+  }
+
+  exec(['--word', word, '--goto'])
+    .then((result) => {
+      const path = `${getCurrentWorkingDirectory()}/${result.output}`;
+      atom.open({ pathsToOpen: [path], newWindow: false });
+
+      if (result.messages) {
+        atom.notifications.addSuccess(result.messages);
```
trotzig: I think this will probably say "ImportJS: imported `foo`" (which is a bug). It's probably good to display messages here though, because we might want to show something in the future.

atom/lib/import-js.js
```diff
@@ -0,0 +1,302 @@
...
+        return;
+      }
+
+      const editor = atom.workspace.getActiveTextEditor();
+      if (editor.getText() !== result.output) {
+        editor.buffer.setTextViaDiff(result.output);
+      }
+
+      if (result.messages) {
+        // TODO improve the format of this potential wall of text
```
trotzig: Haha, yeah. We should improve message handling overall some day.

trotzig:
Looks good except for not applying `--word` after selecting something (see inline comment). Feel free to address and merge at your own convenience. 

Thanks!

atom/.eslintrc.js
lencioni: It basically enforces the [Airbnb JS style guide](https://github.com/airbnb/javascript), which is very popular. It seemed like a good choice because it is the one that most people will be familiar with. If you want to see specifically which rules, look in these files:

- https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/rules/legacy.js
- https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/rules/es6.js

atom/keymaps/import-js.cson
lencioni: Just ship I guess. Maybe I should remove the comment.

atom/lib/AskForSelectionView.js
lencioni: I don't think it is a dealbreaker, but I see two advantages to using paths over indexes:

1. In the off chance that the order changes during the import operation (e.g. adding, moving, or removing a file), an index will produce unexpected results.
2. Using an index requires that we resolve the paths for each word a second time. This probably makes it slower than it needs to be.

atom/lib/getEnv.js
lencioni: When you log in to a shell, there's a bunch of stuff that happens. Some of these things end up setting or modifying environment variables which are important for commands to be executed as you would expect. This consistentEnv package (which is used by atom linter) will get the correct ENV, consistently.

I originally wrote this so that the results would be cached, but I since realized that the package does this internally so this is unnecessary. I can simplify it some.

atom/lib/import-js.js
lencioni: I think I did, but since I didn't have it configured it gave me non-relative paths that were all wrong, so I had to do some manual stuff. It would be nice if we finished #186 so that things would work better out of the box.

atom/lib/import-js.js
lencioni: They could if Node had destructuring or if we used Babel or something to transpile. Node 6 is supposed to include destructuring.

atom/lib/getEnv.js
lencioni: Also, I don't think we will need this after porting the core to JS.

atom/lib/import-js.js
lencioni: Ah, I forgot about disable line!

atom/lib/import-js.js
lencioni: I just tried it out and I don't see a bug. Since the selections contains the words that need to be imported and their indexes in the list, I don't see why `--word` is needed. And actually, instead of calling `importWord` with the selections, we should really just call a new function `importSelections`.

atom/lib/import-js.js
lencioni: I haven't seen it show anything here.

lencioni:
@trotzig the package manager is built in to Atom itself. https://atom.io/packages

I don't really know how to add a package to there.

atom/lib/import-js.js
```diff
@@ -166,100 +166,84 @@ function askForSelections(selectionsToAskFor, previousSelections) {
...
 
 /**
  * @param {String} selections
  */
-function fixImports(selections) {
-  const args = [];
-  if (selections) {
-    args.push('--selections', selections);
+function importSelections(selections) {
+  exec(['--selections', selections])
```
trotzig: This still needs `--word`. Did you see [my comment on an earlier commit](https://github.com/trotzig/import-js/pull/231#issuecomment-197041695) about that?

lencioni: Are you sure? Here's [my response to your comment](https://github.com/trotzig/import-js/pull/231#discussion_r56357271).

trotzig: Sorry, missed that comment. I'll responding in that thread now.

atom/lib/import-js.js
trotzig: It's true that we shouldn't need the `--word`, because the words are already there. But that's not how it currently works. This probably works because the eslint stuff kicks in, finds your unimported word and imports it (using the selected index).

lencioni: Alrighty then! I pushed a new commit that addresses this.

trotzig:
Looks good! Thanks for addressing my concern. Merge at your own convenience! Let me know if you need help publishing anything (the npm/apm package perhaps?).

PR #241, opened 2016-04-30
Remove "ImportJS:" prefix from messages

lencioni:
This is leftover from when we only had a Vim plugin. Now that we support
4 editors and hope to support more, it is clear that each editor may
have a different way of displaying this type of information, so we want
to leave this up to each editor plugin to determine what is best.


trotzig:
Good call! I don't even think plugins will want to add any prefix tbh.

Issue #118, opened 2016-01-09
Change default declaration keyword to import


lencioni:
ES2015 modules (ESM) is the future and we can embrace this by changing the default declaration keyword from `var` to `import`. What do you say?

trotzig:
Absolutely! We just have to make sure to mention it in the release notes. Also, this should be a semver minor version change, agree?

lencioni:
Sure, why not?

trotzig:
Alright, I'm about to take this on now. In celebration of your birthday so to speak. :)

lencioni:
\o/

PR #242, opened 2016-04-30
Use relative paths by default

### LINKED ISSUE #186: Use relative paths by default

lencioni:
Similar to using import as the default declaration keyword (#118), I
think we should use relative paths by default since that is the standard
way of doing things. Using non-relative paths requires a tool like
webpack to be configured in specific ways and should also probably
require additional configuration for ImportJS.

Fixes #186

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,7 +307,7 @@ an empty array `[]` to avoid stripping out extensions.
 
 ### `use_relative_paths`
 
-This option is disabled by default. By enabling it, imports will be resolved
+This option is enabled by default. When enabled, imports will be resolved
 relative to the current file being edited.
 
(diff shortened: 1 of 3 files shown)


trotzig:
👍

trotzig:
I'll merge this, assuming that the next release we ship is 1.0.0. If we end up having to push an interim version, we can just bump the minor number to 0.8.0 with this included.

PR #245, opened 2016-04-30
Messaging v2

trotzig:
I added a few more commits to improve messaging, plus a bugfix to prevent having to "resolve" variables for which no modules match.

### DIFF
--- a/lib/Importer.js
+++ b/lib/Importer.js
@@ -53,7 +53,7 @@ class Importer {
   import(variableName) {
     const jsModule = this.findOneJsModule(variableName);
     if (!jsModule) {
-      this.message(`No JS module to import for variable \`${variableName}\``);
+      this.message(`No JS module to import for \`${variableName}\``);
       return this.results();
     }
 
@@ -86,6 +86,7 @@ class Importer {
       // The current word is not mappable to one of the JS modules that we
       // found. This can happen if the user does not select one from the list.
       // We have nothing to go to, so we return early.
+      this.message(`No JS module found for \`${variableName}\``);
       return this.results();
     }
 
--- a/lib/Importer.js
+++ b/lib/Importer.js
@@ -145,7 +145,7 @@ class Importer {
     });
 
     if (!Object.keys(unusedVariables).length && !undefinedVariables.size) {
-      // return early to avoid rewriting imports
+      // return early to avoid unnecessary computation
       return this.results();
     }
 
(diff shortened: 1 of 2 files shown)


trotzig:
I'm merging this because I'm about to continue working on things that might conflict.

lib/__tests__/Importer-test.js
```diff
@@ -104,7 +104,7 @@ describe('Importer', () => {
       it('displays a message', () => {
         subject();
         expect(result.messages[0]).toMatch(
-          RegExp(`No JS module to import for variable \`${word}\``));
+          RegExp(`No JS module to import for \`${word}\``));
```
lencioni: This isn't really a regexp, is it? Why aren't we using toEqual here?

lib/Importer.js
```diff
@@ -154,24 +154,32 @@ class Importer {
     const unusedImportVariables = new Set();
     Object.keys(unusedVariables).forEach((variable) => {
       const lineNumbers = unusedVariables[variable];
-      if (anyNumbersWithinRange(lineNumbers, oldImports.range)) {
+      if (oldImports.range.length > 1 &&
```
lencioni: I read this post about not using `>` and thought it was an interesting idea. http://llewellynfalco.blogspot.com/2016/02/dont-use-greater-than-sign-in.html

lencioni: Also, I'm not really sure I understand why we need this, even after reading your commit message.

lencioni:
👍 thanks for making these improvements!

lib/__tests__/Importer-test.js
trotzig: Yeah, we should. I'll update this and other places. I think we did it in the past because we used to show timings.

trotzig: 676bfc17e807e82e95bf6929d123e251421c9793

lib/Importer.js
trotzig: If the buffer looks like this:

```
var span = <span/>;
```

`oldImports.range` will be `[0]`, even though there are no imports. Which led us to think that `span` was an import, even though it wasn't. There's probably a better fix for this, because I realize that this won't work if the file consists of only one import, and nothing else.

```
const foo = require('foo');
```

lencioni: Yeah, that's what I was worried about.

PR #248, opened 2016-05-01
Break down findJsModulesFor()

lencioni:
This function was too big and unwieldy. By breaking it into smaller functions, we can more easily see what is going on.


trotzig:
👍

PR #250, opened 2016-05-01
Use export default instead of module.exports

lencioni:
I find this syntax to be a little nicer than the module.exports. Now
that we have a babel plugin that enables this, we might as well use it.

### DIFF
--- a/lib/xregexp.js
+++ b/lib/xregexp.js
@@ -4,7 +4,7 @@ const originalXRegExp = require('xregexp');
  * Wrapper for xRegExp to work around
  * https://github.com/slevithan/xregexp/issues/130
  */
-function xRegExp(regexp, flags) {
+export default function xRegExp(regexp, flags) {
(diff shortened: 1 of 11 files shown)


lib/xregexp.js
```diff
@@ -4,7 +4,7 @@ const originalXRegExp = require('xregexp');
  * Wrapper for xRegExp to work around
  * https://github.com/slevithan/xregexp/issues/130
  */
-function xRegExp(regexp, flags) {
+export default function xRegExp(regexp, flags) {
```
trotzig: Unrelated, but I think we should be able to update xregexp and delete this wrapper file.

trotzig:
👍 Merge at will.

lib/xregexp.js
lencioni: I think that bugfix is in 3.1.1 which hasn't been released yet.

Issue #235, opened 2016-03-21
Node env configuration should use `const` as declaration keyword by default


lencioni:
At least until node gets `import`s.

lencioni:
Node will support `import` once it lands in V8 (https://bugs.chromium.org/p/v8/issues/detail?id=1569) and when the loader spec is finalized (http://whatwg.github.io/loader/). Some people say maybe in Node v7 (https://github.com/nodejs/help/issues/53#issuecomment-216025299).

PR #251, opened 2016-05-01
Switch declaration keyword from const to import

lencioni:
I used the rewrite feature for this conversion. It missed some in
lib/importjs.js because we had some code above the imports block so I
converted those manually.

### DIFF
--- a/.importjs.json
+++ b/.importjs.json
@@ -1,7 +1,7 @@
 [
   {
     "environments": ["node"],
-    "declaration_keyword": "const",
+    "declaration_keyword": "import",
     "ignore_package_prefixes": ["lodash."],
     "use_relative_paths": true
   },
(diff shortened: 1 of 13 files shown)


.importjs.json
```diff
@@ -1,7 +1,7 @@
 [
   {
     "environments": ["node"],
-    "declaration_keyword": "const",
+    "declaration_keyword": "import",
```
trotzig: you can just remove it since it's the default

lib/__tests__/ImportStatement-test.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
```
trotzig: Do we need the use stricts now?

trotzig:
Yeah!

.importjs.json
lencioni: I considered that, but I didn't because we have talked about defaulting to const for node env. #235

lib/__tests__/ImportStatement-test.js
lencioni: Great question. Now that we are transpiling with babel, maybe not.

PR #253, opened 2016-05-02
Use camelCase for configuration instead of snake_case, add basic configuration validation

lencioni:


lib/Configuration.js
```diff
@@ -102,8 +188,8 @@ export default class Configuration {
    */
   get(key, { fromFile } = {}) {
     return this.configs.find((config) => {
-      const appliesTo = config.applies_to || '**';
-      const appliesFrom = config.applies_from || '**';
+      const appliesTo = config.appliesTo || '**';
+      const appliesFrom = config.appliesFrom || '**';
```
trotzig: we could deconstruct these in the method declaration, right?

`({ appliesTo, appliesFrom }) => {...}`

lib/Importer.js
```diff
@@ -28,7 +28,7 @@ export default class Importer {
     this.editor = new CommandLineEditor(lines);
     this.config = new Configuration(pathToCurrentFile);
     this.pathToCurrentFile = pathToCurrentFile;
-    this.messages = [];
+    this.messages = Array.from(this.config.messages);
```
trotzig: Nice way to solve the issue of getting warnings to clients.

trotzig:
Thanks!

trotzig:
I'm going to check the travis output to make sure that tests were actually run. See https://github.com/Galooshi/import-js/pull/257 for context.

trotzig:
Yeah, it looks like tests haven't been run. :(

lib/Configuration.js
lencioni: Yeah, and we could probably also provide defaults

```js
({ appliesTo = '**', appliesFrom = '**' }) => {...}
```

lencioni: Oh, but we actually use the `config` object later. We can still simplify this a bit with some destructuring in the function body. I'll add a commit.

lencioni:
I've rebased and plan to merge this once Travis is green.

PR #257, opened 2016-05-02
Make test fail if linting fails

trotzig:
Due to how we were using a semicolon to combine two scripts, a failure
to the first command wouldn't cause the entire run to fail. Changing to
&& will make everything fail if the first command fails.


lencioni:
I had used the semicolon when our code wasn't passing linting at all. I think it is probably okay to change it to this now. Looks like we need to add eslint-plugin-import to devDependencies. I'll take care of that.

PR #262, opened 2016-05-07
Speed up tests

trotzig:


lib/__tests__/findMatchingFiles-test.js
```diff
@@ -0,0 +1,108 @@
...
+        mkdirp.sync(path.dirname(fullPath));
+        fs.closeSync(fs.openSync(fullPath, 'w')); // create empty file
+      });
+
+      return findMatchingFiles(lookupPath, word);
+    };
+  });
+
+  afterEach(() => {
+    childProcess.execSync(`rm -r ${tmpDir}`);
```
lencioni: What do you think about using rimraf instead of this? That should give us windows compatibility, for instance. https://github.com/isaacs/rimraf

trotzig: Sure thing, although I'd suggest pushing that until we add proper windows support.

lencioni: Might as well switch now since we are already using the mkdirp package here, and rimraf in our npm scripts.

PR #267, opened 2016-05-14
Unify mocking strategy

lencioni:
We recently made some changes to how we are doing mocking in the
Importer test in order to speed it up. To make things consistent, I am
applying these techniques everywhere.


trotzig:
Looks good. Did you notice any performance improvements as the result of this change?

lencioni:
I didn't pay very close attention since these tests were already pretty fast. Jest is parallelized already, so it won't matter as long as we have one slow test.

PR #268, opened 2016-05-14
Add Flow

lencioni:
I'd like to use a static type checker to make our code mode readable and
to help us find bugs before they happen. It seems that our options are
currently Flow or TypeScript. I don't have experience with either
really, so I decided to give Flow a whirl. Since we are already partly
in the Facebook ecosystem with Jest, it seems like this might be a good
bet.

This adds some flow annotations to a few of our files, to see how it
feels. It ended up finding one possible bug already, which is a nice
sign. If we like the direction this is heading, we can expand it to all
of our files, and then eventually remove the `// @flow` pragma and just
have it on automatically.

One somewhat unfortunate downside I ran into was the need to add a Babel
transform for class properties, since Flow requires annotations on class
properties and this is the way it achieves that. This is only
unfortunate because this hasn't been solidified enough in the spec to
make it to stage 3, so it may end up changing dramatically or never
making it into the language officially, which might mean that we'll end
up with some tech debt here. However, I think the tradeoff is worth it
in this case and since we are already going in on a non-standard Flow
syntax, it seems like we might as well do this too.

### DIFF
--- a/lib/findJsModulesFor.js
+++ b/lib/findJsModulesFor.js
@@ -1,3 +1,5 @@
+// @flow
+
 import escapeRegExp from 'lodash.escaperegexp';
 import minimatch from 'minimatch';
 import sortBy from 'lodash.sortby';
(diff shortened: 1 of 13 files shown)


lencioni:
@trotzig I'd love your thoughts on this. If you are into it, we should probably convert the remaining files ASAP.

lib/findJsModulesFor.js
```diff
@@ -7,57 +9,52 @@ import findMatchingFiles from './findMatchingFiles';
...
- */
-function findImportsFromEnvironment(config, variableName) {
+function findImportsFromEnvironment(
+  config: Object, // TODO change Object to Configuration
+  variableName: string
+): Array<JsModule> {
   return config.environmentCoreModules()
-    .filter(dep => dep.toLowerCase() === variableName.toLowerCase())
-    .map(dep => new JsModule({ importPath: dep }));
+    .filter((dep: string): boolean => dep.toLowerCase() === variableName.toLowerCase())
```
trotzig: This is a little chatty, but I understand why it has to be.

lib/formattedToRegex.js
```diff
@@ -1,15 +1,14 @@
...
  *   snake_case
  *   camelCase
  *   PascalCase
  * and turns it into a regex that you can use to find matching files.
- *
- * @param {String} string
- * @return {String}
  */
-export default function formattedToRegex(string) {
+export default function formattedToRegex(string: string): string {
```
trotzig: cool: cool): cool

lib/resolveImportPathAndMain.js
```diff
@@ -31,6 +24,10 @@ function resolveForPackage(filePath) {
 
   let mainFile = json.main;
   const match = filePath.match(/(.*)\/package\.json/);
+  if (!match) {
+    return [null, null];
+  }
+
```
trotzig: Is this the bug?

trotzig:
I'm okay with this, as long as you can guide me along when I need help. The biggest downside as I see it is that it will be slightly more difficult for outsiders to contribute. We have a few layers that not everyone might be familiar with now.

It's funny though that I've been so happy lately that I'm not coding in Java anymore. And now I just reviewed code that looked just like ... Java... ☕

lib/resolveImportPathAndMain.js
lencioni: Yep

lencioni:
> The biggest downside as I see it is that it will be slightly more difficult for outsiders to contribute. We have a few layers that not everyone might be familiar with now.

Although the unfamiliarity aspect is true, I hope that the type checking will make it easier to contribute once that hurdle has been crossed.

PR #269, opened 2016-05-15
Add more Flow annotations

lencioni:
While working on this, I ran into what looks like a bug to me in
Flow.

  https://github.com/facebook/flow/issues/1789

I modified Configuration#constructor to work around this for now.

### DIFF
--- a/lib/ImportStatements.js
+++ b/lib/ImportStatements.js
@@ -1,8 +1,13 @@
+// @flow
+
 // Class that sorts ImportStatements as they are pushed in
 import flattenDeep from 'lodash.flattendeep';
 import partition from 'lodash.partition';
 import sortBy from 'lodash.sortby';
 import uniqBy from 'lodash.uniqby';
+
+import Configuration from './Configuration';
+import ImportStatement from './ImportStatement';
 
 const STYLE_IMPORT = 'import';
 const STYLE_CONST = 'const';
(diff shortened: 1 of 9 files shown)


lib/ImportStatements.js
```diff
@@ -1,9 +1,14 @@
+// @flow
+
 // Class that sorts ImportStatements as they are pushed in
 import flattenDeep from 'lodash.flattendeep';
 import partition from 'lodash.partition';
 import sortBy from 'lodash.sortby';
 import uniqBy from 'lodash.uniqby';
 
+import Configuration from './Configuration';
+import ImportStatement from './ImportStatement';
```
trotzig: I like how this makes dependencies more obvious.

lib/Importer.js
```diff
@@ -329,62 +310,55 @@ export default class Importer {
...
+        matchingImportStatement.namedImports.indexOf(variableName) !== -1;
+    }
+
+    const matchedModule = new JsModule({
+      importPath: matchingImportStatement.path,
+      hasNamedExports,
+    });
+
+    return matchedModule;
   }
```
trotzig: Thanks for making this method better. It was a little hard to follow along with the diff though.

lib/ImportStatements.js
lencioni: Yeah, me too!

lib/Importer.js
lencioni: Yeah, I should have broken that up into separate commits.

PR #275, opened 2016-05-20
Use `require.resolve` to improve `goto`

### LINKED ISSUE #274: "goto module" not working for deep link into node_module
I have a line `import withGlobal from 'mocha-wrap/withGlobal';`. This is `withGlobal.js` inside `node_modules/mocha-wrap`.

However, in 0.7.4 in sublime, when I "goto module", it takes me to an empty file in "$PROJECT_ROOT/mocha-wrap/withGlobal". I suspect the fix to #273 may have broken this use case?

trotzig:
This fixes #274.

When you try to go to a module that reside inside a package dependency,
we wouldn't correctly resolve that to the right file path. Example from
the bug report:

  import withGlobal from 'mocha-wrap/withGlobal';

`mocha-wrap` is a package dependency, and the import path should resolve
to `node_modules/mocha-wrap/withGlobal.js`.

By replacing our custom package dependency resolution strategy with
`require.resolve`, we make sure that the right file path is opened.

To make this change easier to test, and to provide better error
handling, I decided to wrap `require.resolve` in our own
`requireResolve` module.

Apart from adding the tests, I've also verified that this works by
trying to go to `shallowCompare` with this import statement:

  const shallowCompare = require('react/lib/shallowCompare');

Before this change, goto would resolve to
`$project_root/react/lib/shallowCompare``. After the change, it's
correctly resolved to
`$project_root/node_modules/react/lib/shallowCompare.js`

### DIFF
--- a/lib/JsModule.js
+++ b/lib/JsModule.js
@@ -6,6 +6,7 @@ import escapeRegExp from 'lodash.escaperegexp';
 
 import Configuration from './Configuration';
 import ImportStatement from './ImportStatement';
+import requireResolve from './requireResolve';
 import resolveImportPathAndMain from './resolveImportPathAndMain';
 
 function normalizePath(pathToNormalize: string): string {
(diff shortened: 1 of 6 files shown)


lib/JsModule.js
```diff
@@ -164,15 +165,9 @@ export default class JsModule {
...
-    // This is likely an alias that points to a package, so let's try to find
-    // its main file from its package.json file.
-    const filePath = `node_modules/${this.importPath}/package.json`;
-    const main = resolveImportPathAndMain(filePath, [])[1];
-    if (main) {
-      return `node_modules/${this.importPath}/${main}`;
-    }
-
-    return this.importPath;
+    // If all of the above fails to find a path, we fall back to using
```
lencioni: I wonder if there is more in here that we can simplify like this.

lencioni:
Thanks for fixing this!

lib/JsModule.js
trotzig: I think there is. I didn't think too broadly about it though, my time was limited for the fix.

Issue #271, opened 2016-05-18
Remove state and side-effects in Importer


lencioni:
We'd like to break up this class into smaller modules to make this code more testable and maintainable. I looked into doing this this morning, and I think before we do that we need to remove most state and side-effects in Importer. Once everything is a pure function, it will be much easier to move things around.

PR #278, opened 2016-05-22
Move some things around

lencioni:
Partly to start on #271, I decided to move some easy targets off of their classes. This makes these functions truly private and moves us slightly toward a more functional approach which I think will make our code clearer, easier to move around, and easier to test.

### DIFF
--- a/lib/Configuration.js
+++ b/lib/Configuration.js
@@ -173,10 +173,6 @@ function checkCurrentVersion(minimumVersion: string) {
   );
 }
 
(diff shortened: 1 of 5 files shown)


trotzig:
Looks good! I felt so-so about inlining flow type annotations because it ends up making the functions quite big, but I'm okay with it. 

Also, this will get merged knowing about jest issues with Node 6. More info: https://github.com/facebook/jest/issues/1048

PR #281, opened 2016-05-22
Fix local config

trotzig:


lib/__mocks__/requireResolve.js
```diff
@@ -1,7 +1,7 @@
-let resolvedPaths;
+let resolvedPaths = {};
 
-function __setResolvedPaths(paths: object) {
-  resolvedPaths = paths;
+function __addResolvedPath(alias: string, path: string) {
```
lencioni: I think this is fine for now, because of the way our Importer test is structured, but in the long run having to trace up every level to get enough context to understand a given test makes the tests a lot less readable--I think we should try to keep test setup more localized going forward.

lencioni:
> I'm hoping we can move away from lookupPaths soon anyway.

Can you expand on this?

lencioni:
This PR seems to have some linter errors.

trotzig:
> I'm hoping we can move away from lookupPaths soon anyway.

With relative paths, there really is no reason to use lookup paths. Also, I think `require.resolve` won't find things if you for instance use webpack with custom lookup paths. I guess we could do is feed all lookup paths to `require.resolve`. Not sure if we can completely move off of them though, that might have been a naive statement.

lib/__mocks__/requireResolve.js
trotzig: Agreed.

Issue #99, opened 2016-01-05
Running the Sublime plugin in Windows


trotzig:
The ImportJS package in Package Control is only available to OSX users. I choose to make that limitation because I don't have a Linux or Windows environment available to test things in at the moment. 

@janpaul123, @shreeya, @kevinkehl, @lencioni: Does anyone have another environment than OSX where they would be willing to test import-js? I'd mainly like to see how the Sublime plugin works, but I guess it would be good to test the Emacs and vim plugins too if possible.

kevinkehl:
Emacs plugin will work fine in Linux. Windows will work as well as using the shell works in Windows (meaning I have no idea if it will work or not). I don't have a Windows machine, but I'll try using the dev VMs they provide.

vini175pa:
I'm using Linux Mint 64 Bit with Sublime Text 3083. And, apart of that problem i'm having (almost fixing it), it works pretty well.
I'll test it on Windows 10!

trotzig:
Hey @vini175pa, that's probably why you didn't find it through Package Control. I only added osx to begin with, as I didn't have an opportunity to test other platforms. 

I've added Linux to the mix in this PR: https://github.com/wbond/package_control_channel/pull/5103

trotzig:
I've done some initial testing on windows. The main blocker is that we rely on `find` and `egrep` when searching for js files matching a certain string. There are no direct equivalents on Windows (unless you install cygwin), so the `import-js` call fails. Initially, import-js searched files using `Dir.glob` directly in Ruby code. The switch to external `find` and `egrep` commands was done because it improved performance significantly. Though part of that performance win was because we used to look through the `node_modules` folder. Since we no longer do that, we might be able to get away with `Dir.glob` again.

lencioni:
If `Dir.glob` is still too slow, we might want to consider doing OS detection and choosing whatever is best for the given OS. [This is the approach that Overcommit takes](https://github.com/brigade/overcommit/blob/282f569df/lib/overcommit/os.rb), [for instance](https://github.com/brigade/overcommit/blob/de5f0eb8cd/lib/overcommit/utils/file_utils.rb#L22-L25).

lencioni:
https://github.com/facebook/jest/blob/052af961fb469f78ddcbd623b39622416fc21da9/packages/jest-haste-map/src/crawlers/node.js

trotzig:
That looks like a good approach.

lencioni:
Fixed by 0994bc3.

PR #282, opened 2016-05-28
Windows support

trotzig:


lib/__tests__/findMatchingFiles-test.js
```diff
@@ -6,103 +6,121 @@ import rimraf from 'rimraf';
...
-    word = 'foo';
-    existingFiles = [];
-    lookupPath = tmpDir;
-
-    subject = () => {
-      existingFiles.forEach((file) => {
-        const fullPath = `${tmpDir}/${file}`;
-        mkdirp.sync(path.dirname(fullPath));
-        fs.closeSync(fs.openSync(fullPath, 'w')); // create empty file
+['node', 'find'].forEach((finder) => {
```
lencioni: I'm not a big fan of dynamic tests, but I understand why you are doing it here.

lib/findMatchingFiles.js
```diff
@@ -45,3 +33,35 @@ export default function findMatchingFiles(
...
+  if (lookupPath === '') {
+    // If lookupPath is an empty string, the `find` command will not work
+    // as desired so we bail early.
+    throw new Error(`lookup path cannot be empty (${lookupPath})`);
+  }
+
+  const formattedVarName = formattedToRegex(variableName);
+  const validFilesRegex = `(/|^)${formattedVarName}(/index)?(/package)?\\.js.*`;
+
+  if (/^win/.test(process.platform)) {
```
lencioni: If we don't see any meaningful performance differences between these methods, we might want to simplify and just use node for this.

lencioni:
I think this fixes #99.

lib/findMatchingFiles.js
trotzig: I'm seeing some pretty significant differences locally. I'll whip up some numbers too when I have time.

lib/__tests__/findMatchingFiles-test.js
trotzig: Yeah. I wanted to have the same test coverage though, without duplicating things. I think this is okay.

PR #283, opened 2016-05-29
Speed up findMatchingFilesWithNode by changing ignore pattern

trotzig:
I was poking around in the source code for `glob` and found that you can
make the ignore pattern more efficient by making it end with globstar
'**' [1]. The same files are ignored, it's just that `glob` will avoid
traversing the whole `node_modules` folder (which it seems to have done
before).

This took importjs calls down from 1.4s to 0.7s for the following
command (run from the import-js repo root):

  importjs word glob test.js

[1]: https://github.com/isaacs/node-glob/blob/68f2509d237babec64cf03adf72531127fd80339/common.js#L218

### DIFF
--- a/lib/findMatchingFiles.js
+++ b/lib/findMatchingFiles.js
@@ -39,7 +39,7 @@ function findMatchingFilesWithNode(
   validFilesRegex: string
 ): Array<string> {
   return glob.sync(`${lookupPath}/**/*.js*`, {
-    ignore: './node_modules/**/*',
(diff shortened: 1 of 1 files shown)


trotzig:
@lencioni it would be interesting to see if this makes any difference for you. I added a way to test the node finder with an env variable, so you should be able to do something like this:

```bash
time IMPORT_JS_USE_NODE_FINDER=true importjs word foo path/to/file.js
```

PR #287, opened 2016-06-25
Faster find

trotzig:
This PR does a few things to make the process of importing faster:
- `find`ing files is now handled asynchronously (and in parallell). This speeds up the `fix` command when you are importing more than one variable.
- Introduce a `importjsd` daemon process that you can communicate with using stdin and stdout. This speeds up every command since we no longer have to shell out and spin up a node environment for every command. 
- Integrate with [Watchman](https://facebook.github.io/watchman/). This is only used when run in combination with the daemon (because there's a startup cost), and speeds up finding files (by a lot!)

I've been preparing the vim and sublime plugins (vim:[faster-find](https://github.com/Galooshi/vim-import-js/commits/faster-find), sublime:[faster-find](https://github.com/Galooshi/sublime-import-js/commits/faster-find)) so that they're ready for the changes. I was thinking we can release this and let it marinate in the wild a little first though before we merge the branches on the editor plugins. The old command-line tool will continue to work the same way as before, so we can merge this separately if we want to.


trotzig:
@lencioni: Would you mind having a look at this when you have a chance? No rush, this doesn't have to go out quickly. 

We'll have to update the Atom plugin at some point too, perhaps you could look into that? I think @kevinkehl might be interested in this as well for the emacs plugin. 

To run this locally, I've got the `faster-find` branch checked out for import-js, and then that's `npm link`ed. Then I have the `faster-find` branch checked out for `vim-import-js` as well, and that folder is symlinked in my pathogen bundles folder. 

Debugging is harder now that we're dealing with a background process. The daemon will log to a file, and you can set the log level in `.importjs.json` (`"logLevel": "debug"`). All daemon process log to the same place, and the path to that file is printed on daemon startup. Now, editors will usually hide this output, so you usually have to run `importjsd` manually to understand where things get logged. For me, this is what I see on startup:

```bash
⨠ importjsd
DAEMON active. Logs will go to: /var/folders/1l/_t6tm7195nd53936tsvh2pcr0000gn/T/importjs.log
```

trotzig:
The basic idea for how to run this in our editor plugins is simple:

Plugins start the daemon by running `importjsd --parent-pid <pid>`. The parent pid thing is optional but without it the daemon can't itself check to see if it has become a zombie.

Plugins write commands to stdin of the daemon process. Something like 
```json
{
  "command": "word",
  "commandArg": "bar",
  "fileContent": "const foo = bar()\n",
  "pathToFile": "lib/foo.js",
}
```

Plugins read output from stdout of the daemon process. The output will be JSON strings (one-line), with the same structure as for the old `importjs` command-line tool.

Plugins should kill the daemon on exit, although there's a mechanism that will auto-kill daemons that don't have a running parent (checks every 30s). This is where the `--parent-pid` comes into play.

trotzig:
A short note about the list of commits: The work I've done has been fairly exploratory. I've tried to make commits clear about what they're doing, but I haven't worried about making sure that the code is in a great state in between every commits. That's why you'll see the build breaking from time to time, and commits leaving things half-done.

bin/importjs.js
```diff
@@ -1,21 +1,4 @@
 #!/usr/bin/env node
 
-/* eslint-disable strict */
-
-'use strict';
```
lencioni: Do we not actually need this anywhere anymore? I think removing it might end up breaking older node, but maybe that's okay.

lib/fix-node-path.js
```diff
@@ -0,0 +1,14 @@
...
+//
+// This is because eslint will look for modules relative to where it installed.
+// The eslint we are using is local to import-js, so any plugin referenced for
+// the file we are linting will have to be looked up relative to that file.
+//
+// Technique from http://stackoverflow.com/questions/11969175
+const oldPath = process.env.NODE_PATH;
+process.env.NODE_PATH = `${oldPath || ''}:${process.cwd()}/node_modules/`;
+require('module').Module._initPaths(); // eslint-disable-line no-underscore-dangle
+
```
lencioni: nit: killable blank line

lib/JsModule.js
```diff
@@ -44,13 +46,15 @@ export default class JsModule {
     relativeFilePath,
     stripFileExtensions,
     stripFromPath,
+    variableName,
```
lencioni: Why did you move this into class state?

lib/daemon.js
```diff
@@ -2,20 +2,29 @@ import readline from 'readline';
...
 
-  rlInterface.on('line', (jsonCommand: string) => {
-    const command = JSON.parse(jsonCommand);
+  rlInterface.on('line', (jsonPayload: string) => {
+    const payload = JSON.parse(jsonPayload);
     const importer = new Importer(
-      command.fileContent.split('\n'), command.pathToFile);
+      payload.fileContent.split('\n'), payload.pathToFile);
 
     // TODO: don't automatically invoke function here because of security issues
```
lencioni: At this point it should be pretty trivial to just throw an error if `!functionName`

lib/daemon.js
```diff
@@ -28,7 +28,9 @@ export default function daemon() {
       process.stdout.write(`${JSON.stringify(result)}\n`);
     }).catch((error) => {
       // TODO: print entire stack trace here
-      process.stderr.write(error.toString());
+      process.stdout.write(JSON.stringify({
```
lencioni: I'm not so sure about this, but we'll see where this PR goes.

lib/daemon.js
```diff
@@ -10,27 +11,43 @@ const commandsToFunctionNames = {
   add: 'addImports',
 };
 
+function log(message) {
+  fs.appendFile('/tmp/importjsd.log', `${message}\n`, (err) => {
+    if (err) {
+      throw new Error(err);
```
lencioni: Isn't `err` an `Error`? Could this be `if (err) throw err;`?

lib/WatchmanFileCache.js
```diff
@@ -0,0 +1,122 @@
...
+
+const CACHE = new Set();
+
+// Flag to quickly check if the cache is enabled.
+let enabled = false;
+
+function subscribe({ client, fbWatch, relativePath }): Promise {
+  const subscription = {
+    // Match any `.js` file
+    expression: ['allof', ['match', '*.js*']],
```
lencioni: This should maybe include `['type', 'f']`

Instead of the glob matcher, I wonder if we can use `suffix` and if that would be better. Maybe this works? `['suffix', ['js', 'jsx']]`

lencioni: re `type`, we might also need to include symlinks. https://facebook.github.io/watchman/docs/expr/type.html

lib/WatchmanFileCache.js
```diff
@@ -0,0 +1,122 @@
...
+      subscribe({
+        client,
+        fbWatch: resp.watch,
+        relativePath: resp.relativePath,
+      }).then(resolve).catch(reject);
+    });
+  });
+}
+
+export default {
```
lencioni: Why use an object here instead of a `class`?

lencioni: Or actually, since there is no state, object is fine. Alternatively you could export the functions separately, but I'm not sure that gives us anything.

lib/findMatchingFiles.js
```diff
@@ -49,6 +51,26 @@ function findMatchingFilesWithNode(
   });
 }
 
+function findMatchingFilesWithWatchman(
+  lookupPath: string,
+  validFilesRegex: string
+): Promise<Array<string>> {
+  const tester = new RegExp(validFilesRegex, 'i');
```
lencioni: `tester` is not a very good variable name.

lib/findMatchingFiles.js
```diff
@@ -4,8 +4,9 @@ import childProcess from 'child_process';
 
 import glob from 'glob';
 
-import formattedToRegex from './formattedToRegex';
 import WatchmanFileCache from './WatchmanFileCache';
+import formattedToRegex from './formattedToRegex';
+import normalizePath from './normalizePath';
```
lencioni: Was this added here by accident?

lib/normalizePath.js
```diff
@@ -1,10 +1,12 @@
 import escapeRegExp from 'lodash.escaperegexp';
 
+const CWD_PATTERN = RegExp(`^${escapeRegExp(process.cwd())}`);
```
lencioni: This should be a small perf win as well.

lib/__tests__/findMatchingFiles-test.js
```diff
@@ -64,8 +58,11 @@ import findMatchingFiles from '../findMatchingFiles';
         ];
       });
 
-      it('returns an empty array', () => {
-        expect(subject()).toEqual([]);
+      it('returns an empty array', (done) => {
```
lencioni: I'm not sure if Jest/Jasmine has this feature but in Mocha, you can return a promise in your test for a similar effect, FYI.

lib/__tests__/findMatchingFiles-test.js
```diff
@@ -1,12 +1,14 @@
+/* global spyOn */
```
lencioni: We should just add this to the .eslintrc file for tests.

lencioni: Actually I'm a little surprised that the `jest` `env` doesn't give us this. Might be worth looking into that.

lib/daemon.js
```diff
@@ -1,10 +1,30 @@
...
+/* eslint-disable no-console */
+console.log = (...args) => {
+  logFile.write(`${util.format.apply(null, args)}\n`);
+};
+console.trace = console.log;
+console.debug = console.log;
+console.info = console.log;
+console.warn = console.log;
+console.error = console.log;
+/* eslint-enable no-console */
```
lencioni: It might be nice to put this console redirection in its own file.

.importjs.json
```diff
@@ -2,6 +2,7 @@
   {
     "environments": ["node"],
     "ignorePackagePrefixes": ["lodash."],
+    "logLevel": "debug",
```
lencioni: I remember thinking user-specific config would have been helpful for another thing, but I don't remember what it was anymore. Can you think of other things that user-specific config would be useful for?

lib/daemon.js
```diff
@@ -41,6 +42,7 @@ export default function daemon() {
   const config = new Configuration('importjsd');
 
   loglevel.setLevel(config.get('logLevel'));
+  loglevelMessagePrefix(loglevel);
```
lencioni: It might be nice to be explicit with our options

README.md
```diff
@@ -493,11 +496,17 @@ files before adding the `-exec` part.
...
-To speed up importing, import-js can run as a daemon. Start the daemon by
-running `importjsd`. The daemon accepts commands sent via `stdin`. Each command
-is a (oneline) JSON string ending with a newline. The command structure is
-basically the same as for the command-line tool, but wrapped in JSON instead of
-expressed on the command line. Here are a few examples:
+Instead of invoking the `importjs` command-line tool every time you import
+something, you can run ImportJS in a background process and communicate with
+it using `stdin` and `stdout`. This will make importing faster because we don't
+have to spin up a node environment on every invocation. Chances are your editor
+plugin is already using the daemon for importing.
```
lencioni: It might be a good idea to be more clear about how if you are using one of the official editor plugins, that you won't need to do anything to get these benefits. This section is intended for advanced usage only.

lib/daemon.js
```diff
@@ -36,21 +37,38 @@ const commandsToFunctionNames = {
...
+
 export default function daemon() {
   // The `importjsd` here is mostly a dummy file because config relies on a
   // `pathToCurrentFile`. Normally, this is the javascript file you are editing.
   const config = new Configuration('importjsd');
 
   loglevel.setLevel(config.get('logLevel'));
   loglevelMessagePrefix(loglevel);
-  originalConsoleLog(`DAEMON active. Logs will go to: \n${pathToLogFile}`);
+  originalConsoleLog(`DAEMON active. Logs will go to: ${pathToLogFile}`);
```
lencioni: It might be nice to throw in `ImportJS ${VERSION}` at the beginning of this as well.

lib/daemon.js
```diff
@@ -36,21 +37,38 @@ const commandsToFunctionNames = {
   add: 'addImports',
 };
 
+commander.option('--parent-pid <n>', parseInt).parse(process.argv);
```
lencioni: Should we include the parent PID as a prefix in the logs?

lencioni:
👍

lib/WatchmanFileCache.js
trotzig: Cool, will look into this. I mostly had this copied from some example somewhere I think.

bin/importjs.js
trotzig: Oh, didn't think of that. I'll make sure to test it.

lib/JsModule.js
trotzig: I needed to be able to call `toImportStatement` in a context where I didn't have access to the variable name.

lib/daemon.js
trotzig: True.

lib/daemon.js
trotzig: If I remember correctly, I needed this to make communicating with the daemon in vim simpler. It also makes communication in the sublime plugin simpler, because I can wait for a line to be written to stdout.

lib/daemon.js
trotzig: Yeah probably. This goes away later anyway.

lib/WatchmanFileCache.js
trotzig: I mean, there is state. The `CACHE` is state, plus `enabled`. This whole file could use some love, plus some tests.

lib/__tests__/findMatchingFiles-test.js
trotzig: Yeah, I was surprised too. I'm guessing it's because they want you to use mocking instead.

lib/__tests__/findMatchingFiles-test.js
trotzig: Ah, it looks like you can (sort of): https://facebook.github.io/jest/docs/tutorial-async.html

lib/daemon.js
trotzig: yeah, I guess.

lib/normalizePath.js
```diff
@@ -0,0 +1,12 @@
+import escapeRegExp from 'lodash.escaperegexp';
+
+export default function normalizePath(rawPath: string): string {
+  if (!rawPath) {
+    return './';
+  }
+  const normalized = rawPath.replace(RegExp(`^${escapeRegExp(process.cwd())}`), '.');
```
trotzig: Yep, that might be a good idea.

lib/findMatchingFiles.js
trotzig: Yeah, I think it snuck in early. I use it later.

lib/normalizePath.js
trotzig: Yep. I was afraid that it would lock us down to not being able to change cwd mid-process. But after looking at this more, we can't really switch cwd anyway. Mostly because of the `fix-node-path` thing, which can only happen on initialization.

.importjs.json
trotzig: I think this is the first config option that I've come across where it made sense.

README.md
trotzig: Yeah, I'll add something about that. Good call.

lib/daemon.js
trotzig: Good idea!

lib/daemon.js
trotzig: Haha, I just (30s ago) wrote this in another comment. We should.

lib/findMatchingFiles.js
lencioni: I think `validFilesRegex` should be `validFilesPattern`, and for this `RegExp` instance to be `validFileRegex`

bin/importjs.js
trotzig: I just tried this with Node 4.4.4, and it's working.

lib/fix-node-path.js
trotzig: ba432d6

lib/daemon.js
trotzig: I think this actually makes sense (printing to stdout). It makes the contract with the clients simpler, since they can expect a single line of JSON at all times. 

Ouch! Just realized there's no newline character at the end here. I'll add that if I hadn't already in a later commit.

trotzig: Already dealt with apparently. 👍

lib/daemon.js
trotzig: 9424d7c

lib/WatchmanFileCache.js
trotzig: d061f98 switches to `suffix`. I'll look into `type` next.

trotzig: I've played around with `type` now, and I can't get it to work for symlinks. And having directories end up in the cache isn't really a problem, since we apply a fairly sophisticated regexp to grep for the "right" files. And the other finders are not filtering out directories either, so doing so here would make things inconsistent. I'm going to leave this untouched.

lib/findMatchingFiles.js
trotzig: e404b28

README.md
trotzig: 07ecfa3

lib/daemon.js
trotzig: 8a75d50

lib/daemon.js
trotzig: a803e9c

trotzig:
I think I've addressed all of the comments that you had @lencioni. I'm going to go ahead and merge. My plan is to publish a new version (1.0.0) with the changes. And then I'll let that marinate for a bit before I go ahead and push new versions of plugins.

Issue #293, opened 2016-06-30
Fix Meteor package imports when useRelativePaths is true


rhettlivingston:
Meteor uses the convention of prepending "meteor/" to Meteor package names to signal to the build environment that the reference is to a Meteor package as opposed to an NPM package. This conflicts with the existing strategy invoked by the useRelativePaths=true option (the default). Instead of the desired

```js
import { Meteor } from 'meteor/meteor';
```

importjs is outputting

```js
import { Meteor } from './meteor/meteor';
```

An exception similar to the one already present for node modules probably needs to be added at https://github.com/Galooshi/import-js/blob/a8a3ce020cd650a4993d2bff14ce6e953eea7c3b/lib/Configuration.js#L271 to achieve the desired result.

This issue was created as a result of discussion in #289.

rhettlivingston:
Simply adding the line referring to 'meteor/' below at lib/Configuration.js#L271 seems to fix the problem.
```js
    if (this.get('useRelativePaths', { fromFile: relativeFilePath }) &&
      !relativeFilePath.startsWith('meteor/') &&
      !relativeFilePath.startsWith('node_modules')) {
```
All tests passed with this addition and it fixed a test in my work for #289 that was failing due to the problem. So, I have some confidence in its effectiveness.

Are there tests that are explicitly targeted towards verifying the effectiveness of the above-noted line containing 'node_modules' that I could duplicate and adjust to more explicitly verify the effectiveness of the 'meteor/' line? If not, are we good without more checks or do I need to put one somewhere?

trotzig:
I think that's good enough. If you look at a8a3ce02 (the commit that introduces the `node_modules` exception), there's [a test that you can use as a template](https://github.com/Galooshi/import-js/commit/a8a3ce02#diff-da30310350dcd91444d26504857a03c0R1252). It's in `Importer_test.js` which is a bit of a pain to work with due to its size. We have a long-standing plan to break it up, but we haven't prioritized it yet.

rhettlivingston:
This issue was resolved by the changes in 04cc6c2 and e343494 which were merged to master via 629e49b.

As noted before, I need to work on my commit messages. Had to go read the github docs too. :frowning:

Issue #289, opened 2016-06-28
Environments Entry for "meteor"


rhettlivingston:
I second @awatson1978 in saying "thank you, thank you for putting this together"!

To complement #288, I am investigating other aspects of getting this package to play really nice with Meteor.

Though node.js is a part of meteor under the hood, what we see as our "core" modules is quite different. As a start to get the groupImports feature to play nicely with Meteor, I'd like to request an addition to the supported "environments". I am currently vetting my list of what I believe are Meteor's "core" modules on their forum, but believe it will be close to the following:

```js
  meteor: [
    'meteor/accounts-base',
    'meteor/blaze',
    'meteor/check',
    'meteor/ddp-client',
    'meteor/ddp-rate-limiter',
    'meteor/ejson',
    'meteor/email',
    'meteor/http',
    'meteor/check',
    'meteor/meteor',
    'meteor/mongo',
    'meteor/random',
    'meteor/reactive-var',
    'meteor/session',
    'meteor/templating',
    'meteor/tracker',
  ],
```

Would you be willing to accept a PR for this?

Also, I have several other ideas on the burner for things like 
 - automatically identifying our 3rd party packages (that middle group in groupImports),
 - easing our extensive use of namedExports (unlike NPM, our package system defines the exports so we have the leg up you were discussing in #107 for this one),
 - exposing babel-lint's "allowImportExportEverywhere" option to us (we'll need to turn it on),
 - and handling the common usage of the import for side-effects statement in Meteor (`import './thisModulesName.http;'` and `import './thisModulesName.css';`).

As well as thoughts about the configuration processing (love the idea of accepting a js file and wonder if that could be expanded on a touch to allow plugins that specify things like above for different environments). Should I approach these via more "issues" or a more general offline discussion followed by more issues?

trotzig:
Hi @rhettlivingston, and thanks for the ❤️!

Let's keep the discussion here to begin with, and then we can split it out into separate issues later on. I'll try to give a few thoughts on your initial post. 

> Would you be willing to accept a PR for this? (meteor core modules)

Yes, absolutely.

> automatically identifying our 3rd party packages (that middle group in groupImports),

How are these dependencies defined? (I'm not a Meteor user myself)

> easing our extensive use of namedExports (unlike NPM, our package system defines the exports so we have the leg up you were discussing in #107 for this one),

Great! I think you/we have an opportunity to lead the effort forward here. Perhaps we can get more packages to adopt the Meteor style here. 

[...]

PR #297, opened 2016-07-01
Define meteor environment

rhettlivingston:
These changes should fully address #289 "Environments entry for Meteor" and "#293 Fix Meteor package imports when useRelativePaths is true".

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,14 +176,18 @@ import styles from './bar.scss';
 ### `environments`
 
 This list of environments controls what core modules are available when
-importing. The only supported value right now is `["node"]`, and if you use it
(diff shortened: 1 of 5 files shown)


trotzig:
Solid -- Thanks for contributing!

If I could ask for one thing for future commits, it would be to flesh out the commit messages a little. Both @lencioni and I are fans of verbose commit messages (@lencioni has a good blog post about why we care about it: https://medium.com/brigade-engineering/the-secrets-to-great-commit-messages-106fc0a92a25#.g2bjmrrpa).

rhettlivingston:
You're welcome! And thy wish on commit messages shall be fulfilled.

PR #301, opened 2016-07-07
Javascript configuration

trotzig:


trotzig:
@rhettlivingston - can you have a look at this and see if it's enough to help move the Meteor support forward? I didn't use a 3rd party library mostly because of the legacy (now deprecated) local configuration support. We can reconsider replacing our own implementation when we have removed the local config support.

rhettlivingston:
Good timing. Just finished a couple of days of addressing an issue with eslint-plugins-import that was effecting Meteor users. I'll try to take a look at this tonight.

README.md
```diff
@@ -450,6 +457,58 @@ When using `appliesFrom` only a subset of configurations are supported:
...
+such case, the function is invoked with the following arguments:
+
+- `pathToCurrentFile`: (always available) A path to the file you are editing.
+- `pathToImportedModule` (not available for some options) A path to the
+  file/module you are importing.
+
+Here's an example of how to dynamically control the `declarationKeyword`
+configuration option based on the file you are importing:
+
+```.js
```
lencioni: You have an extra `.` here

README.md
```diff
@@ -450,6 +457,58 @@ When using `appliesFrom` only a subset of configurations are supported:
...
+    if (isTestFile(pathToImportedModule)) {
+      return 'const';
+    }
+    return 'import';
+  }
+}
+```
+
+Here's a more elaborate example taking both `pathToImportedModule` and
+`pathToCurrentFile`into account:
```
lencioni: missing space before "into"

lencioni:
LGTM

README.md
trotzig: Thanks. Fixed in 29be3fe

README.md
trotzig: Thanks. Fixed in 29be3fe

trotzig:
I think what could be missing from this PR is a bit of error handling. What if the config file can't be parsed for instance? Let me know if you think it's needed. In such case I might push a different PR later.

lencioni:
It would also be nice to have a deprecation warning for the deprecated configuration style, so people know to update.

rhettlivingston:
The error handling is a definite concern. As a new user working with the json file, I lost a lot of time battling simple mistakes.

I'm working to fully understand the code to try to answer the "will it work for Meteor" question. The answer is certainly not less than "its a big step in the right direction".

trotzig:
@rhettlivingston, @lencioni: I added a few commits to improve error handling. If you're okay with it, I'll merge this so that we can continue to work on top of it.

Going forward, I think we could cut out some complexity and make the whole process better by using a 3rd party dependency for the configuration, but it's probably not worth doing that until we've cut out the support for local configuration (through `appliesTo`/`appliesFrom`).

rhettlivingston:
No objections. We'll only know for sure as we start using it. I hope to get into this again after a day of yard work tomorrow (assuming I survive mowing in 100 degree weather :smile:).

trotzig:
💪 🌲 ☀️

trotzig:
The master branch started failing after merging this. https://travis-ci.org/Galooshi/import-js/builds/143416013

I'm pretty sure it has nothing to do with these changes. Instead, it's something that was updated in eslint, and due to how we aren't locking our dependencies to a specific version, changes in 3rd party libraries can cause our build to suddenly fail. We could look into start using npm-shrinkwrap here, although my history with that tool isn't completely free from annoyances.

I'll look into the failing build later on (probably tomorrow).

lencioni:
> We could look into start using npm-shrinkwrap here

npm shrinkwrap makes sense for deployable apps, not for packages like this one.

trotzig:
i guess I see why, but wouldn't at least locking versions be a good thing? Though I guess that leads to people having to download more versions of stuff than what they need.

lencioni:
Theoretically, semver should protect us.

lencioni:
Locally I uninstalled eslint and reinstalled it, and that did not reproduce the issue. I was able to reproduce by clearing out node_modules and installing everything. This tells me that the problem isn't with eslint, but somewhere else. I should have run `npm outdated` before clearing things out, but I wasn't thinking. If you still have a clean set of node modules, it might be nice to dig into this a bit more. I'll keep looking here.

lencioni:
It seems somehow related to the `no-shadow` rule and our flow definitions. I'm going to disable that rule until we get this sorted out.

lencioni:
I disabled the rule in c2ab6fa65cd84f50a73a655f1b8952088f982ca6

trotzig:
Thanks @lencioni! Unfortunately, I did the same thing as you did last night (installed latest eslint, watch things still work, wiped node_modules, installed everything again, get error).

Issue #290, opened 2016-06-29
Allow environments to specify full set of configuration options


trotzig:
See https://github.com/Galooshi/import-js/issues/289 for context. The idea is to allow e.g. the `meteor` environment to specify aliases, package dependencies, etc instead of only having a list of core modules.

Issue #292, opened 2016-06-29
Move environment core modules into configuration


trotzig:
As discussed in https://github.com/Galooshi/import-js/issues/289. This is depending on https://github.com/Galooshi/import-js/issues/290.

Issue #199, opened 2016-02-19
Auto-resolve named exports from ES2015 modules


lencioni:
It is a common pattern to have a module that exports multiple things, and possibly no default thing. Example:

``` js
export function foo() {
  ...
}

export function bar() {
  ...
}
```

Now, if I want to use one of these functions, I am left in the dust by import-js. It would be nice if we could resolve these named exports automatically.

lencioni:
Short of parsing the entire project into an AST or something, I think that we might be able to solve for the most common cases using some regexes. And, I think that a partial solution that we can iterate on is better than no solution here.

trotzig:
This is related to #184 and #183. I've been thinking about this and I think we could add a cache of some sort to the file system. `.import-js.cache.json` or something. Depending on how long it takes for this cache file to generate, we can either add an explicit command (e.g. `import-js init`) or make it a one-time startup thing. And if we're smart about what we store in this cache, we can probably use a little bit of time on some import calls to check if we need to update the cache. 

I have a hard time seeing a solution that doesn't use a cache. 

When you say "it is common", how common do you mean? I haven't come across many projects that do that.

lencioni:
In the project that I am working in, `ag "export function"` shows up 149 times. I would expect this to become more common as ES2015 modules becomes more widely used.

jscheid:
Hi, this would be great to have!

> Short of parsing the entire project into an AST

Just a heads up that this should be fairly easy with Babel, see https://github.com/babel/babel/tree/master/packages/babel-core#usage

Obviously you'd want to use something like https://github.com/mikeal/watch to keep it snappy.

I'm pretty busy at the moment but I'll try and see if I can whip up a proof of concept, or let me know if there's a better way I can help move this forward. Cheers!

lencioni:
Thanks! We are more concerned with performance than how difficult it is to do, which is why I suggested the regex approach. If you can come up with a method that can do this for a large project without introducing friction for getting import-js set up and without slowing it down too much, that would be really great. 

We are also in the middle of a rewrite from Ruby to JavaScript. If you plan on hacking, you should probably start from the convert-to-js branch.

trotzig:
Yeah, it would be awesome to have more people involved! As @lencioni said, we're currently rewriting the whole project in JavaScript. I wouldn't say we're in the middle, we're almost ready to merge the convert-to-js branch.

jscheid:
[...]

PR #302, opened 2016-07-09
Move env core modules to configuration (#290)

trotzig:
This will allow environments to specify more things than just core
modules. We need this to fully implement Meteor support (epic story
covered in #289).

Since you can specify multiple environments to inherit from, we need to
merge some configuration options. Most importantly, we need to merge
coreModules. We probably also want to merge aliases and namedModules.
Even though we currently only use coreModules in our environment
configs, I decided to add support for merging aliases and namedModules
as well.

This is a little bit more complicated than it needs to be. That's mostly
because we're still supporting local configuration. Once we move off of
that, we can simplify things.


trotzig:
@rhettlivingston, @lencioni: Let me know what you think of this. I'm interested to hear your thoughts on my decision to merge certain config options, and how I ended up implementing that.

trotzig:
This ended up fixing #292 too I think.

lib/Configuration.js
```diff
@@ -7,6 +7,8 @@ import semver from 'semver';
 
 import FileUtils from './FileUtils';
 import JsModule from './JsModule';
+import meteor from './environments/meteor';
+import node from './environments/node';
```
lencioni: I think it would be better if these files were called `meteorEnvironment` and `nodeEnvironment` for maximum readability.

lib/Configuration.js
```diff
@@ -34,68 +37,11 @@ const DEFAULT_CONFIG = {
...
-    'tls',
-    'tty',
-    'url',
-    'util',
-    'v8',
-    'vm',
-    'zlib',
-  ],
-};
+const MERGABLE_CONFIG_OPTIONS = [
```
lencioni: At this point in the code, I don't really understand what "mergable" means. Maybe it is worth adding a comment?

lib/Configuration.js
```diff
@@ -263,15 +221,39 @@ export default class Configuration {
...
+      fromFile
+    );
+  }
 
-    if (typeof value === 'function') {
-      return value({
-        pathToCurrentFile: this.pathToCurrentFile,
-        pathToImportedModule: fromFile,
-      });
+  mergedValue(values: Array<any>, key: string, fromFile: string): any {
```
lencioni: This function barely uses `this`. What do you think about moving it to the top level and passing in `pathToCurrentFile` instead? That would make it a pure function.

lib/Configuration.js
```diff
@@ -263,15 +221,39 @@ export default class Configuration {
...
+    let mergedResult;
+    for (let i = 0; i < values.length; i++) {
+      let value = values[i];
+      if (typeof value === 'function') {
+        value = value({
+          pathToCurrentFile: this.pathToCurrentFile,
+          pathToImportedModule: fromFile,
+        });
+      }
+      if (!MERGABLE_CONFIG_OPTIONS.includes(key)) {
```
lencioni: This call should maybe be moved out of the loop.

lib/__tests__/Configuration-test.js
```diff
@@ -363,14 +363,97 @@ describe('Configuration', () => {
...
+        path.join(process.cwd(), '.importjs.json'),
+        [
+          {
+            aliases: {
+              foo: 'foosball',
+            },
+          },
+          {
+            aliases: {
+              bar: 'barsketball',
```
lencioni: see also BASEketball https://en.wikipedia.org/wiki/BASEketball

lib/__tests__/Configuration-test.js
```diff
@@ -363,14 +363,97 @@ describe('Configuration', () => {
...
+            },
+          },
+          {
+            aliases: () => ({
+              baz: 'bazeball',
```
lencioni: see also Brännboll https://en.wikipedia.org/wiki/Br%C3%A4nnboll

lib/__tests__/Configuration-test.js
```diff
@@ -363,14 +363,97 @@ describe('Configuration', () => {
...
+        path.join(process.cwd(), '.importjs.json'),
+        [
+          {
+            namedExports: {
+              foo: ['foosball'],
+            },
+          },
+          {
+            namedExports: () => ({
+              bar: ['barsketball'],
```
lencioni: What happens if I have `foo ['something else']` here? Will it override the earlier `foo` or get added to it?

lib/__tests__/Configuration-test.js
```diff
@@ -355,6 +355,103 @@ describe('Configuration', () => {
...
+        bar: 'barsketball',
+        baz: 'bazeball',
+      });
+    });
+  });
+
+  describe('#get namedExports', () => {
+    beforeEach(() => {
+      // TODO: replace the array-style configuration here when we have an
+      // environment that specify namedExports.
```
lencioni: I don't think I really understand these TODOs.

lencioni:
LGTM

lib/Configuration.js
```diff
@@ -263,15 +221,39 @@ export default class Configuration {
...
+    let mergedResult;
+    for (let i = 0; i < values.length; i++) {
+      let value = values[i];
+      if (typeof value === 'function') {
+        value = value({
+          pathToCurrentFile: this.pathToCurrentFile,
+          pathToImportedModule: fromFile,
+        });
+      }
+      if (MERGABLE_CONFIG_OPTIONS.indexOf(key) === -1) {
```
lencioni: We could bring in a polyfill for this pretty easily. We could grab airbnb-js-shims and be set for a bunch of things, or we can bring it in directly from https://github.com/ljharb/array-includes

rhettlivingston:
This is awesome! Yes, I believe you've got #292 covered also. I survived the mowing and will be digging into this tomorrow.

Below is the base list of namedExports that Meteor needs to complete the picture provided by those coreModules. They can be thrown in that nifty new environments file.
```
  namedExports: {
    'meteor/accounts-base': ['AccountsClient', 'Accounts', 'AccountsServer'],
    'meteor/blaze': ['Blaze'],
    'meteor/check': ['check', 'Match'],
    'meteor/ddp-client': ['DDP'],
    'meteor/ddp-rate-limiter': ['DDPRateLimiter'],
    'meteor/ejson': ['EJSON'],
    'meteor/email': ['Email'],
    'meteor/http': ['HTTP'],
    'meteor/meteor': ['Meteor'],
    'meteor/mongo': ['Mongo'],
    'meteor/random': ['Random'],
    'meteor/reactive-var': ['ReactiveVar'],
    'meteor/session': ['Session'],
    'meteor/templating': ['Template'],
    'meteor/tracker': ['Tracker'],
  },
```

rhettlivingston:
Another important aspect of my thoughts for #290 involves the existing packageDependencies routine. 

packageDependencies is currently very Node specific. Thus, it seems it _should_ be part of the Node environnment. However, that might be breaking because I don't think your users have to specify the Node environment to get it right now. 

Regardless, I think the packageDependencies need to be pulled from the configuration just like namedExports. Meteor, needs both the existing packageDependencies (because it is based on Node and does now have a package.json and all that comes with it) and a routine to scan Meteor's package system.

With support for packageDependencies in the configuration, that can just be implemented there...

One question,,, I'd like to take advantage of your existing routine for scanning a module to find its named exports in Meteor's packageDependencies routine. So, will I be able to get to that from within environments/meteor.js?

lib/Configuration.js
trotzig: Fair. I'll update.

trotzig: eeef564

lib/Configuration.js
trotzig: I added a comment in 8c4044e.

lib/Configuration.js
trotzig: I guess that makes sense. Fixed in bd160be.

lib/Configuration.js
trotzig: I guess, though then I'd have to potentially invoke the function in two places. I'll leave it here.

lib/__tests__/Configuration-test.js
trotzig: Fantastic.

lib/__tests__/Configuration-test.js
trotzig: Now we're talking.

lib/__tests__/Configuration-test.js
trotzig: Yeah, I was on the fence about these. The reasoning is that once we no longer have local configuration through an array, we can't test configuration merging this way. The meteor config will soon have a namedExports config, and at that time we should start using it instead of this workaround.

lib/__tests__/Configuration-test.js
trotzig: Yes. I thought about adding support for merging those arrays but I don't know if it's worth it.

trotzig:
@rhettlivingston: I added `namedExports` in d6340b4. 

For `packageDependencies`, we can move the default `package.json` scanning into a separate module, then make the default configuration (not the `node` one) simply use this module. And then the `meteor` environment file can override `packageDependencies` and return its own list (making use of that same module to scan `package.json`). 

There is no routine to scan for named exports. We've talked about adding one in #199, but it hasn't been implemented - mainly because of performance concerns. We haven't come up with a solution that's fast and robust enough. It should be noted however that the discussion in that issue started when this whole project was written in Ruby, and importing was done through making a call to the `importjs` CLI tool. Since then we've converted to javascript and implemented a daemon running in the background. There's a lot more room for long-running processes now, and indexing/caching etc. Also, #199 discusses adding a scanner for all js files, not just package dependencies. 

If you start working on moving `packageDependencies` to configuration, leave a line here. Otherwise I might start doing it tonight or tomorrow night.

trotzig:
> packageDependencies is currently very Node specific. Thus, it seems it should be part of the Node environnment. However, that might be breaking because I don't think your users have to specify the Node environment to get it right now.

I agree that's specific to Node, although our `node` environment is more for "pure" node libraries (like ImportJS for instance). For projects meant to be run in a browser, you shouldn't use the `node` environment. But you might still need `package.json`. Does that make sense?

rhettlivingston:
>I added namedExports in d6340b4.

Thank you! Already pulled it.

>For packageDependencies, we can move the default package.json scanning into a separate module, then make the default configuration (not the node one) simply use this module. And then the meteor environment file can override packageDependencies and return its own list (making use of that same module to scan package.json).

That sounds good.

>There is no routine to scan for named exports. We've talked about adding one in #199, but it hasn't been implemented - mainly because of performance concerns. We haven't come up with a solution that's fast and robust enough. It should be noted however that the discussion in that issue started when this whole project was written in Ruby, and importing was done through making a call to the importjs CLI tool. Since then we've converted to javascript and implemented a daemon running in the background. There's a lot more room for long-running processes now, and indexing/caching etc. Also, #199 discusses adding a scanner for all js files, not just package dependencies.

Wow, I need more sleep. Yes, I somehow crossed up your code for digging out import statements with what it would take to dig out export statements. 

As you allude, my initial case is not a full scanner. It will be limited in that I will know exactly what files in a package to go to and the file will often be a trivial one included specifically to gather all of the exports from a more complicated package into a common point. So, I think the performance impact will be limited.

So perhaps I'll create the routine to scan a module for exports. It should have a lot of similarity to your findCurrentImports / ImportStatement code for extracting imports.

>For projects meant to be run in a browser, you shouldn't use the node environment. But you might still need package.json. Does that make sense?

Actually, very much so. Meteor is an isomorphic environment. A lot of our code runs on both the client and the server because we have speculative execution on the client. So I already benefit from having this code where it is and have not found a reason to bring in the Node coreModules yet.

>If you start working on moving packageDependencies to configuration, leave a line here. Otherwise I might start doing it tonight or tomorrow night.

I'm not jumping at this aspect yet, so feel free to go for it.

PR #303, opened 2016-07-10
Move packageDependencies to configuration

trotzig:
This will allow us to implement custom behavior for meteor.

### DIFF
--- a/lib/Configuration.js
+++ b/lib/Configuration.js
@@ -36,6 +36,10 @@ const DEFAULT_CONFIG = {
   stripFromPath: null,
   tab: '  ',
   useRelativePaths: true,
+  packageDependencies: ({ config }: Object): Array<string> =>
+    findPackageDependencies(
+      config.workingDirectory,
+      config.get('importDevDependencies')),
 };
 
 // Default configuration options, and options inherited from environment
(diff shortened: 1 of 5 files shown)


trotzig:
Ping @rhettlivingston, @lencioni

lib/Configuration.js
```diff
@@ -36,6 +36,10 @@ const DEFAULT_CONFIG = {
   stripFromPath: null,
   tab: '  ',
   useRelativePaths: true,
+  packageDependencies: ({ config }: Object): Array<string> =>
```
lencioni: It might be nice to be more specific about the shape of this Object with this type declaration. We could likely do this better in a number of places.

lib/Configuration.js
```diff
@@ -259,14 +231,45 @@ export default class Configuration {
...
+    fromFile: string
+  ): any {
+    let mergedResult;
+    for (let i = 0; i < values.length; i++) {
+      let value = values[i];
+      if (typeof value === 'function') {
+        value = value({
+          pathToCurrentFile: this.pathToCurrentFile,
+          pathToImportedModule: fromFile,
+          config: this,
```
lencioni: If you wanted, this function could still be a pure function with minimal fuss if you passed in `config` instead of `pathToCurrentFile`. But at this point it is starting to feel a bit potato potato.

lencioni: Still, it might be worth doing, to keep the Configuration API slim and to keep state to a minimum.

lencioni:
LGTM

rhettlivingston:
Doesn't `packageDependencies` need to be in the KNOWN_CONFIGURATION_OPTIONS?

trotzig:
> Doesn't packageDependencies need to be in the KNOWN_CONFIGURATION_OPTIONS?

I guess, though I don't see this option being used in user config. That's partly why I didn't document it either.

rhettlivingston:
Ahhh. I was testing it that way. I can change my approach.

rhettlivingston:
LGTM

lib/Configuration.js
trotzig: I'll hold off on bouncing this back to a regular function again. But yeah, I see where you're coming from.

lib/Configuration.js
trotzig: I'll leave this for a later time, I hope that's okay with you.

Issue #305, opened 2016-07-12
useRelativePaths = false causes local modules to be grouped with packages


rhettlivingston:
With useRelativePaths = true, import-js creates the following imports in one of my modules
```js
import { DDPRateLimiter } from 'meteor/ddp-rate-limiter';
import { Meteor } from 'meteor/meteor';

import { SimpleSchema } from 'meteor/aldeed:simple-schema';
import { ValidatedMethod } from 'meteor/mdg:validated-method';
import { _ } from 'meteor/underscore';

import { Lists } from '../lists/lists';
import { Todos } from './todos';
```

When I set useRelativePaths = false, the two local modules moved into the packages section
```js
import { DDPRateLimiter } from 'meteor/ddp-rate-limiter';
import { Meteor } from 'meteor/meteor';

import { Lists } from 'imports/api/lists/lists';
import { SimpleSchema } from 'meteor/aldeed:simple-schema';
import { Todos } from 'imports/api/todos/todos';
import { ValidatedMethod } from 'meteor/mdg:validated-method';
import { _ } from 'meteor/underscore';
```

Note that I am working on the packageDependencies-is-configurable branch though I believe this is independent of that. 

Do I correctly remember that there is a point in the code that looks at the leading "." in the path to determine whether it is dealing with a package?

trotzig:
I think the bug here is that they are never treated as package dependencies. Regular relative imports end up in a different group than regular absolute imports. I think that's what's causing the first example to look "correct". 

There's a check in lib/ImportStatements.js that you might have to adjust for meteor: https://github.com/Galooshi/import-js/blob/7d8dfbff30fe5bb3809fd1c8b78a6dce1705bc83/lib/ImportStatements.js#L69

trotzig:
Sorry, didn't mean to close.

rhettlivingston:
Yes, you're correct. I misinterpreted what was happening.

rhettlivingston:
And now I was too hasty. :smile: 

You were correct in the reason it was happening at that time. But now that I've got the packageDependencies properly specified, I still have the problem for the right reason.

You were also correct in that the aforementioned line in lib/ImportStatements.js needs an adjustment for Meteor.

I think the best fix is to replace 

```js
  // If there is a slash in the path, remove that and everything after it.
  // This is so that imports for modules inside package dependencies end up
  // in the right group (PATH_TYPE_PACKAGE).
[...]

PR #306, opened 2016-07-12
Recognize 'meteor/somepackage' as a package (#305)

rhettlivingston:
In importStatemetnPathType, the lines
```js
  const path = importStatement.path.replace(/^(.*?)\/.*$/, '$1');
  if (packageDependencies.indexOf(path) !== -1) {
```
strip a meteor package path such as 'meteor/aldeed:simple-schema' down to nothing but 'meteor' and then fail to match the package name contained in the packageDependencies array.

The regex line was placed there to allow imports for modules inside packages to be properly placed in the package group.

Rather than change the regex to skip the 'meteor/' prefix, I have chosen in this change to eliminate the need for the regex by matching the packageDepencies against either the whole path or the start of the path up to a path divider as shown below.
```js
  if (packageDependencies.some((pkg: string): boolean =>
        importStatement.path === pkg ||
        importStatement.path.startsWith(`${pkg}/`))) {
```
This should fix the problem in a fashion that is not Meteor specific.

### DIFF
--- a/lib/ImportStatements.js
+++ b/lib/ImportStatements.js
@@ -63,11 +63,13 @@ function importStatementPathType(
     return PATH_TYPE_CORE_MODULE;
   }
 
-  // If there is a slash in the path, remove that and everything after it.
-  // This is so that imports for modules inside package dependencies end up
-  // in the right group (PATH_TYPE_PACKAGE).
-  const path = importStatement.path.replace(/^(.*?)\/.*$/, '$1');
-  if (packageDependencies.indexOf(path) !== -1) {
+  // Match if any of the packageDependencies exactly match path or match the
+  // start of the path up to a path divider. This is so that imports for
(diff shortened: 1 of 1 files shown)


lencioni:
Thanks! This also makes the code more readable. 💃

lencioni:
It would be really nice to add a test for this, if you have a moment @rhettlivingston.

rhettlivingston:
I did verify that a test failed if I removed the portion using "startsWith" because I was worried as to whether there was an existing test for recognizing references to modules within npm packages as part of a package. There is.

I'll add a todo for adding a test more specific to my situation of allowing packages to have names containing slashes, like `meteor/aldeed:simple-schema`, to be recognized as packages.

trotzig:
Thanks for making that code more robust and more readable.

PR #307, opened 2016-07-13
Add tests for meteor package recognition (#305)

rhettlivingston:
This commit in response to [lencioni's request](https://github.com/Galooshi/import-js/pull/306#issuecomment-232220379) augments the tests to verify that future changes don't break the fix to package recognition that was added in response to #305.

Two tests were added.

The first verifies that Meteor package dependencies which contain slashes,  "meteor/aldeed:simple-schema", are recognized as packages and separated from non-package dependences.

The second verifies that they are still recognized as packages when the path includes a path to a module inside of the package, i.e. "meteor/aldeed:simple-schema/validate".


lencioni:
Thanks for following up with these tests!

PR #308, opened 2016-07-15
Implement meteor environment's packageDependencies

rhettlivingston:
Meteor includes two package systems, its own and support for NPM packages. Currently, only NPM packages are being recognized as such by the default packageDependencies implementation.

This submission builds on #303 which allows environment configurations to supply a packageDependencies function.

A packageDependencies function is added to environments/meteorEnvironment.js. It returns a list that is a combination of Meteor packages extracted from the .meteor/packages file and NPM packages found via the existing findPackageDependencies function.

### DIFF
--- a/lib/environments/meteorEnvironment.js
+++ b/lib/environments/meteorEnvironment.js
@@ -61,7 +61,7 @@ function meteorPackageDependencies(
     // add 'meteor/' to the start of each name per Meteor convention
     .map((pkg: string) => `meteor/${pkg.trimLeft()}`)
     // eliminate those packages that are considered to be core
-    .filter((pkg: string) => !coreModules.includes(pkg));
+    .filter((pkg: string) => coreModules.indexOf(pkg) === -1);
   return packages;
 }
 
--- a/lib/environments/meteorEnvironment.js
+++ b/lib/environments/meteorEnvironment.js
@@ -29,6 +29,32 @@ function meteorPackageDependencies(
   if (!fs.existsSync(meteorPackagesPath)) {
     return [];
   }
+  // Meteor is an "app" framework. As such, it has both apps and packages. When working with a
+  // module that is part of an app, the list of Meteor packages that a module may import is found in
+  // '.meteor/packages'. This file is actually called a ProjectConstraintsFile in the meteor code.
+  // The internal meteor routine that parses it may be found at
+  // https://github.com/meteor/meteor/blob/f8b1bba606678303ea4e82b8da7c8ae0683b0b7c/tools/project-context.js#L841.
+  // After reverse engineering ProjectConstraintsFile.prototype._readfile at that location, the
+  // following appears to be the pertinent facts for any parser.
+  //  - the only true information within the file is a list of constraints
+  //  - a constraint may not span a line
+  //  - only one constraint may appear on a line
+  //  - a constraint consists of a package name and an optional version constraint separated by
+  //    the '@' symbol
+  //  - white space may not appear within a constraint
+  //  - the '#' symbol signifies that the rest of the line is a comment
+  //  - all white space is ignored
+  //  - package names
+  //      - are allowed to contain [a-z0-9:.\-]
+  //      - must have at least one lowercase letter
+  //      - may not begin or end with a dot or colon
+  //      - may not begin with a hyphen
+  //      - may not contain two consecutive dots
+  //
+  // This routine is only interested in extracting the package names and has no concern for
+  // precisely validating them. An assumption is made that the file is basically valid.
+  // Thus, they may be extracted with a simple global, multiline match of characters allowed
+  // to be in a package name that are at the beginning of a line, possibly following white space.
   const packages = fs.readFileSync(meteorPackagesPath, 'utf8')
     // extract an array of package names from the packages file
     .match(/^\s*([^# @\n]+)/gm)
--- a/lib/environments/meteorEnvironment.js
+++ b/lib/environments/meteorEnvironment.js
@@ -56,10 +56,10 @@ function meteorPackageDependencies(
   // Thus, they may be extracted with a simple global, multiline match of characters allowed
   // to be in a package name that are at the beginning of a line, possibly following white space.
   const packages = fs.readFileSync(meteorPackagesPath, 'utf8')
-    // extract an array of package names from the packages file
-    .match(/^\s*([^# @\n]+)/gm)
+    // extract an array of package names (possibly with preceding whitespace) from the packages file
+    .match(/^\s*[a-z0-9:.\-]+/gm)
     // add 'meteor/' to the start of each name per Meteor convention
-    .map((pkg: string) => `meteor/${pkg}`)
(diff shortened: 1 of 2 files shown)


rhettlivingston:
Tests to follow. Wanted to get it out for comments.

lib/environments/meteorEnvironment.js
```diff
@@ -1,22 +1,46 @@
...
+function meteorPackageDependencies(
+  { config }: Object
+): Array<string> {
+  const meteorPackagesPath = path.join(config.workingDirectory, '.meteor/packages');
+  if (!fs.existsSync(meteorPackagesPath)) {
+    return [];
+  }
+  const packages = fs.readFileSync(meteorPackagesPath, 'utf8')
+    // extract an array of package names from the packages file
+    .match(/^\s*([^# @\n]+)/gm)
```
trotzig: I could be wrong, but I don't think the `m` flag exists for javascript regexps. We've been using https://www.npmjs.com/package/xregexp for that, but I'm not sure you need it here. 

Could packages ever have `@` in their name? It looks like we would ignore them in such case.

lib/environments/meteorEnvironment.js
```diff
@@ -1,22 +1,46 @@
...
+  'meteor/reactive-var',
+  'meteor/session',
+  'meteor/templating',
+  'meteor/tracker',
+];
+
+function meteorPackageDependencies(
+  { config }: Object
+): Array<string> {
+  const meteorPackagesPath = path.join(config.workingDirectory, '.meteor/packages');
```
trotzig: I'm trying to find documentation about what the syntax of this file looks like. Based on the regexp it seems to be a simple text file with a package per line. It could help if there was a comment here explaining things, or something in the commit message.

trotzig:
Looks good to me so far.

lib/environments/meteorEnvironment.js
lencioni: Storybook has an `@` in its package name. https://github.com/kadirahq/react-storybook/blob/master/package.json#L2

rhettlivingston: The `m` flag is definitely working as expected in the code and is needed. The [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#Advanced_searching_with_flags) also includes it.

<package>@<version> is the syntax used in packages to pin a package to a version. I had no need for this particular version.

The best documentation I could find (probably because I trust code more than docs and didn't look extensively at the docs) is the [validatePackageName function](https://github.com/meteor/meteor/blob/devel/packages/package-version-parser/package-version-parser.js#L379) in Meteor's code. It looks for bad characters with the following statement.
```js
  var badChar = packageName.match(/[^a-z0-9:.\-]/);
```
So, "@" isn't allowed.

lib/environments/meteorEnvironment.js
rhettlivingston: I'm looking for the same. May have to just create it and place it here in a comment. "packages" is considered to be an internal file though most users maintain it by hand instead of using the provided `meteor add`, `meteor remove` and `meteor list` commands. It implicitly supports such usage by allowing comments (prefixed with "#") and blank lines.

lib/environments/meteorEnvironment.js
trotzig: My mistake about `m`. I was thinking of the dotall modifier (`s`).

rhettlivingston:
@trotzig Thanks for pushing me to dig out the .meteor/packages syntax. It forced me to enough thought to realize that the code I had was including leading whitespace in the package name instead of stripping it. Comments have been added and corrections made.

rhettlivingston:
While working to create the meteorPackageDependencies function and the meteorNamedExports function I'm noting what is going to become a performance issue.

I'm testing using `importjs fix` and these routines are being called multiple times, on an order related to the number of unique unresolved variables.

This won't matter for the more common `importjs word` or `importjs goto` but it seems like we should seek to avoid it in the case of `importjs fix`.

rhettlivingston:
AFAIK, this is ready to go.

trotzig:
Looks good to me. Thanks!

trotzig:
I didn't look to closely at the performance issue. I see two ways to resolve that: 
- Move the call to `config.get('packageDependencies');` to an earlier stage in the import, and pass the array along to consumers (currently `findJsModulesFor` and `ImportStatements`). 
- Cache the results internally in `meteorEnvironment.js`. We can use `meteorPackagesPath` + last modification time of that file as the cache key. 

I would prefer the first option as it will allow other environments (including the default) to also benefit from not being called multiple times.

rhettlivingston:
The first one sounds best to me for now. But, ultimately, I'm wondering if we'll eventually have to do the second.

My concern with the second is primarily complexity. The namedExports are much more important in this discussion than the packageDependencies. Detecting packageDependency changes that should force a rebuild of the cached value is easy. Detecting namedExport changes involves watching a lot more files. I think that in even a simple Meteor application, it is in the 100s.

So, for now, I think we should punt on building a cache mechanism that can work across multiple import-js operations and simply seek to reduce the actual invocations of the packageDependencies and namedExports functions to once per import-js operation.

But, we will likely evolve.

Note that the Meteor packageDependencies execution is currently a bit naive. It is going to evolve to be a bit module specific. In the Meteor environment, users frequently edit both their app modules and modules within their local packages in the same editing session. Apps and packages, even local packages located within the app directory structure, have different packageDependencies. The packageDependencies implementation I just submitted handles app  dependencies on packages. I've got to add the functionality for detecting Meteor package dependencies on packages.

Basically, I think I'm going to have to start ignoring the current working directory and determining the context of the module import-js is working with. This will involve walking up the directory tree. If I hit a directory containing a Meteor "package.js" file before I hit a directory containing ".meteor/packages", then the module is in a package and will need its packageDependencies determined differently. And, if I find it is under a "node_modules" directory with a package.json located above it, it's a third situation with a third means of determining package dependencies. All three situations could happen in a Meteor project.

lencioni:
There might be another option. We could do a debounced memoization of `config.get('packageDependencies')` so that any call within a threshold returns the cached value.

rhettlivingston:
>There might be another option. We could do a debounced memoization of config.get('packageDependencies') so that any call within a threshold returns the cached value.

I don't know enough about your daemon and batch operations to say for sure, but that sounds like you might be thinking of using the cache from one file to the next.

packageDependencies is definitely going to become a pathToCurrentFile dependent operation (within days hopefully) though there may be many files that get the same answer returned.

Any caching operation that spans files would thus have to use option 2 above where the code within the Meteor environment has the possibility of recognizing the current file's context and returning the particular cached answer for that context.

trotzig:
I think it might be worth implementing time-consuming stuff in a background thread and accept the fact that changes aren't reflected immediately. Doing so probably involves leaning on the Watchman integration a lot more. We can generalize `WatchmanFileCache.js` so that we can hook in arbitrary jobs when files are changed.

Issue #288, opened 2016-06-26
Option to add leading prefix to imports


awatson1978:
Hi there,
First of all, let me just say that this package is such a relief.  Thank you, thank you for putting this together.  I've been in import hell as I get up to speed with imports, and having this package is such a relief already.

So, feature request.

When using Meteor, everything is working swimmingly, but we need an option for a leading prefix.  For instance, I press ``Cmd+Shift+j`` over the word "AddPost", and we get the following situation:

```js
// created import string 
import AddPost from 'imports/ui/workflows/posts/AddPost';

// desired import string 
import AddPost from '/imports/ui/workflows/posts/AddPost';
```

Apparently Meteor want's absolute URLs that begin with a `/`.

Would you be willing to accept a PR for this?

trotzig:
Hi @awatson1978, and thanks for the kind words!

Just to better understand your situation, the leading slash isn't referring to the root of the file system, right? It's a way in Meteor to refer to the root of the app, correct? 

We're definitely open to a PR for this. I think you could hook into [the `toImportStatement` method of the `JsModule` class](https://github.com/Galooshi/import-js/blob/fe7519078dfab6ee3751d2e274c111c4508a8fa3/lib/JsModule.js#L169). That would make all import paths subject to the transformation, which I think is good. 

How would you feel about naming the configuration option `importPathPrefix`? 

You should also be aware that we are considering allowing the `.importjs.json` config file to also be pure javascript (`.importjs.js`). When we do that, we would probably solve this use-case by allowing people to configure a `prepareImportPath` method. But that's part of the long-term plan for import-js, and not something that we'll look into any time soon.

trotzig:
Oh, and I should probably ask before you go down the path of preparing the PR: would relative paths work? I'm looking at a few Meteor examples and it looks like they're using relative paths there. https://www.meteor.com/tutorials/react/components

```js
import AddPost from '../posts/AddPost';
```

awatson1978:
Hi @trotzig,
So, I'm not a core developer on Meteor, although I do maintain a distro for healthcare users. 

As I understand things, the leading slash refers to the root of the _app_.  However, I wouldn't completely rule out the possibility that there's a temp directory somewhere.  The Meteor tool is a pipeline build tool; so it actually compiles into Node code.  Slight chance there's a temp directory involved.  But otherwise, lets just assume that it's the root of the app.   

[...]

PR #310, opened 2016-07-15
Module specifier

trotzig:
This addresses #288. I haven't added anything to the README because I wanted to first know if you think I'm on the right track. 

Also, we're currently at 1000 commits. We should just stop and call the project done. ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,7 @@ The following configuration options can be used.
 - [`ignorePackagePrefixes`](#ignorepackageprefixes)
 - [`minimumVersion`](#minimumversion)
 - [`maxLineLength`](#maxlinelength)
+- [`moduleNameFormatter`](#moduleNameFormatter)
 - [`tab`](#tab)
 - [`logLevel`](#loglevel)
 
@@ -379,7 +380,33 @@ Defaults to `80`. This setting controls when import statements are broken into
 multiple lines.
 
 ```json
-"max-line-length": 70
+"maxLineLength": 70
+```
+
+### `moduleNameFormatter`
+
+Use a function here to control how the resulting module name string will look
+like. It's useful if you for instance want to add a custom prefix to certain
+imports. Apart from the standard `pathToCurrentFile` and `pathToImportedModule`
+values passed in to all configuration functions, this method is also passed a
+`moduleName` value, which in general is what you want to manipulate.
+
+```js
+moduleNameFormatter: ({ moduleName, pathToCurrentFile } => {
+ if (/-test/.test(pathToCurrentFile)) {
+   // Import a mocked version in test files
+   return `mocks/${moduleName}`;
+ }
+
+ if (/^foo/.test(moduleName)) {
+   // Add a leading slash to foo imports
+   return `/${moduleName}`;
+ }
+
+ // Fall back to the original specifier. It's important that this function
+ // always returns a string.
+ return moduleName;
+},
 ```
 
 ### `tab`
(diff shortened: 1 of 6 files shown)


trotzig:
Ping @rhettlivingston, @lencioni (I don't know if you subscribe to PR changes)

lencioni:
Good work!

I'm not convinced about the name `moduleSpecifier`. It's not quite right. Maybe `moduleNameFormatter`?

Also, I'm not sure if there is an advantage to having the original `moduleSpecifier` passed as an argument to the user-defined one. While I like the explicitness of it, I wonder if it would be better to just fall back if the user-defined one returns `null` or `undefined`. However, if we expect to use this pattern in other places and it has advantages there, I can see the case for keeping APIs consistent.

lib/JsModule.js
```diff
@@ -201,7 +201,10 @@ export default class JsModule {
       defaultImport,
       importFunction: config.get('importFunction', { pathToImportedModule }),
       namedImports,
-      path: this.importPath,
+      path: config.get('moduleSpecifier', {
+        pathToImportedModule,
+        moduleSpecifier: this.importPath,
+      }),
```
rhettlivingston: Is it OK that the "path" specified to ImportStatement has been changed from one that might have been useful for actually finding a file to one that is only useful as a ModuleSpecifier in an import statement? If so, should it be called "path"?

rhettlivingston:
> Ping @rhettlivingston, @lencioni (I don't know if you subscribe to PR changes)

I've got a watch on the project for the foreseeable future, so I'm notified on everything.

lib/JsModule.js
trotzig: Good point. I'll rename the variable internally for ImportStatement.

trotzig:
Good, I'll stop the pinging then.

trotzig:
@lencioni I like `moduleNameFormatter` for the option/function and `moduleName` for the variable. I'll change to that. We probably want the original `moduleName` passed in so that you can make tiny adjustments. I thought about falling back if the method returns undefined, but I think it's fine always having to return something.

lib/JsModule.js
trotzig: That turned out to be non-trivial. We use `importPath` in the public API that the editor plugins use, and I'd rather push the rename to a different time as it requires updates to all the plugins as well. I could of course rename it locally only, but I'd rather keep a funky but consistent name than a more correct but inconsistent one.

lencioni: Maybe create an issue with a breaking change label so we don't forget about it?

trotzig: Good call. #311

rhettlivingston:
This looks great to me!

README.md
```diff
@@ -379,7 +380,33 @@ Defaults to `80`. This setting controls when import statements are broken into
...
+### `moduleNameFormatter`
+
+Use a function here to control how the resulting module name string will look
+like. It's useful if you for instance want to add a custom prefix to certain
+imports. Apart from the standard `pathToCurrentFile` and `pathToImportedModule`
+values passed in to all configuration functions, this method is also passed a
+`moduleName` value, which in general is what you want to manipulate.
+
+```js
+moduleNameFormatter: ({ moduleName, pathToCurrentFile } => {
```
rhettlivingston: This is missing a closing parentheses right before the arrow.

trotzig: Thanks! Fixed in d0d386a.

PR #312, opened 2016-07-19
Auto-create namedExports for Meteor packages using metadata

rhettlivingston:
This implements a "90%" solution to automatically creating namedExports for both 3rd-party and local Meteor packages used by an app.

It works for the majority of the packages used today, primarily those still using "api.exports" calls in the package.js file to define their interface to users.

Packages that have already converted to the new api.mainModule calls and do not also use the old api.exports interface are not yet supported. Processing those packages will require the implementation of a routine to scan a module for ES6 exports.

### DIFF
--- a/lib/environments/meteorEnvironment.js
+++ b/lib/environments/meteorEnvironment.js
@@ -78,26 +78,30 @@ function meteorPackageDependencies(
   // precisely validating them. An assumption is made that the file is basically valid.
   // Thus, they may be extracted with a simple global, multiline match of characters allowed
   // to be in a package name that are at the beginning of a line, possibly following white space.
-  const packages: Array<String> = fs.readFileSync(meteorPackagesPath, 'utf8')
+  const packages: Array<string> = fs.readFileSync(meteorPackagesPath, 'utf8')
     // extract an array of package names (possibly with preceding whitespace) from the packages file
     .match(/^\s*[a-z0-9:.\-]+/gm)
     // add 'meteor/' to the start of each name per Meteor convention
-    .map((pkg: String): String => `meteor/${pkg.trimLeft()}`)
+    .map((pkg: string): string => `meteor/${pkg.trimLeft()}`)
     // eliminate those packages that are considered to be core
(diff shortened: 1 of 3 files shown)


lib/environments/meteorEnvironment.js
```diff
@@ -55,35 +78,196 @@ function meteorPackageDependencies(
...
+  //        "testOnly": true
+  //      }
+  //    ],
+  //
+  // From this, we can determine that the namedExports entry should be
+  //
+  //   'aldeed:simple-schema': ['SimpleSchema', 'MongoObject']
+  //
+  // "humanize" was left off because it is testOnly. We may try to get smarter and include that for
+  // some modules in the future.
```
trotzig: Related to this is the `importDevDependencies` configuration option.

lib/environments/meteorEnvironment.js
```diff
@@ -55,35 +78,196 @@ function meteorPackageDependencies(
...
+
+      // TODO: If the "resources" section of the buildIsopack specifies a mainModule, we need to
+      // attempt to scan it to find exports.
+    }
+    // If we found declared exports, create a namedExports entry for them.
+    if (declaredExports.length) {
+      namedExports[`meteor/${pkg}`] = declaredExports;
+    }
+  }
+  return namedExports;
```
trotzig: This method is long, which makes it a little hard to read.

trotzig:
Looks good to me. I left a comment about a long method in meteorEnvironment.js that I think should be addressed just to make things more readable.

lib/environments/meteorEnvironment.js
rhettlivingston: I was on the fence with it myself. I'll break it up.

rhettlivingston:
The long method has been broken up. I'm not sure why it has conflicts. The diff to master on my system is clean.

trotzig:
Thanks! I'll look into the conflict.

trotzig:
All good, thanks for breaking that method up.

PR #313, opened 2016-07-19
Fix Meteor absolute path references to modules

rhettlivingston:
As reported in #288, Meteor wants a '/' in front of absolute paths and interprets them relative to the project root directory.

This PR utilizes the new moduleNameFormatter configuration hook to add the slash in front of module names that are using absolute paths.

This is the final piece required to satisfy #288.

### DIFF
--- a/lib/__tests__/Importer-test.js
+++ b/lib/__tests__/Importer-test.js
@@ -1962,6 +1962,111 @@ foo
         });
       });
 
+      describe('in a meteor environment', () => {
+        beforeEach(() => {
+          configuration.environments = ['meteor'];
+          existingFiles = ['bar/foo.jsx'];
+          text = 'foo';
+          configuration.lookupPaths = ['bar'];
+        });
+
+        describe('with `useRelativePaths=true`', () => {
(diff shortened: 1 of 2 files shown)


trotzig:
Looks good to me! Thanks for following up on top of my changes. I was hoping to get to it myself but I'm glad you did. It would probably have taken me longer to figure out the logic.

rhettlivingston:
You're welcome!

It was pretty quick though I tossed my first implementation. LookupPath's implementation surprised me. The pathToImportedModule was relative to the lookupPath that the imported module had been found on. That means that I really didn't have a "pathToImportedModule" in some cases because there is no way to know what lookupPath it was found on.

Closes #288.

trotzig:
Ah, that's annoying. I wonder if we should pass along `lookupPath` as well, or just make the path include it?

rhettlivingston:
It definitely felt wrong. I don't know what might depend on the path being
the way it is. Passing the particular element of lookupPath that the file
was found in or null if it was found somewhere else would likely help
someone in the future.

PR #314, opened 2016-07-23
Configure Travis to push code coverage to CodeClimate

### LINKED ISSUE #169: Set up code coverage
We recently fixed a bug in goto that would have been caught earlier if we had a test covering it. I think that setting up some code coverage reports would be helpful for us to make this project more resilient by alerting us to gaps that we might want to fill.

lencioni:
We want this project to be resilient and we believe that code coverage
will help us identify gaps in our testing.

The CodeClimate documentation is a little unclear about how to properly
set this up, so I am trying this out. I believe that I also need to set
the CODECLIMATE_REPO_TOKEN environment variable, which I have added
through the Travis interface to prevent this secret from being checked
into code.

  https://travis-ci.org/Galooshi/import-js/settings

I've manually uploaded current code coverage to CodeClimate and it looks
like we are at 93%.

  https://codeclimate.com/github/Galooshi/import-js/coverage

I've also configured CodeClimate to update Pull Requests with an
integration.

Fixes #169.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,11 @@ node_js:
   - "4"
   - "5"
   - "6"
+
+before_script:
+  - npm install -g codeclimate-test-reporter
+
+script: npm run test:cover
+
+after_script:
+  - codeclimate-test-reporter < coverage/lcov.info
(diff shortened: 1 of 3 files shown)


trotzig:
Thanks @lencioni!

PR #315, opened 2016-07-23
Configure CodeClimate to run ESLint

lencioni:
I originally disabled this to help get CodeClimate successfully set up.
Now that we are past this hurdle, let's try enabling this feature.


lencioni:
Unfortunately, it looks like we can't use this because they don't support some of the eslint plugins that we are using. More info: https://github.com/codeclimate/codeclimate-eslint/issues/10

PR #321, opened 2016-07-26
Module side effect imports

trotzig:
This PR has changes from me and @rhettlivingston to add support for side-effect imports. This:

```js
import 'some-module-with-side-effect';
```

A new configuration option, `moduleSideEffectImports` can be used to auto-inject these imports when you run `fix-imports`:

```js
moduleSideEffectImports: ({ pathToCurrentFile }) => {
  if (fooBar(pathToCurrentFile)) {
    return ['index.css', 'messages.json'];
  }
  return [];
},
```


trotzig:
I don't understand the travis error. `lodash` is definitely in the node_modules folder, and `assign` is one of its methods. I'm able to reproduce this locally by generating a fresh node_modules folder, so I think I should be able to track things down.

```
rm -r node_modules
npm install
```

trotzig:
I'm currently thinking that jest might be mocking lodash without being told to. I've tried creating a file importing `lodash` outside of jest and it correctly finds `lodash.assign`. 

Here's a stack trace:

```
TypeError: lodash.assign is not a function
    at Object.ModuleResolver (node_modules/eslint/lib/util/module-resolver.js:42:27)
    at Object.<anonymous> (node_modules/eslint/lib/config/config-file.js:60:16)
    at Runtime._execModule (node_modules/jest-runtime/build/index.js:375:17)
    at Runtime.requireModule (node_modules/jest-runtime/build/index.js:210:14)
    at Runtime.requireModuleOrMock (node_modules/jest-runtime/build/index.js:274:19)
    at Object.<anonymous> (node_modules/eslint/lib/config.js:14:18)
    at Runtime._execModule (node_modules/jest-runtime/build/index.js:375:17)
    at Runtime.requireModule (node_modules/jest-runtime/build/index.js:210:14)
    at Runtime.requireModuleOrMock (node_modules/jest-runtime/build/index.js:274:19)
```

rhettlivingston:
I started seeing this yesterday. I remember being on the verge of locking lodash back, but got sidetracked on another problem. I hadn't figured out the jest relationship. Thanks for finding it!

rhettlivingston:
I've opened a can of worms trying to get the "rewrite" command working. I had a partial solution Friday in which I basically just passed side-effect imports through. Then it started bothering me that the moduleName paths were not being adjusted to conform to the useRelativePaths setting. I'm now considering going with the partial solution and creating issues for some of the shortcomings. As long as we're tolerating the side-effect imports and passing them through, we're at a better place than we were.

The tip of the iceberg I stumbled on is hinted at by the fact that findJsModulesFor and findMatchingFiles are both assuming that we are starting with a variable name. We need the ability to find a JsModule using the module name which should not be manipulated in the same ways and may include a path. That path may be relative to the importing file or to any of the lookupPaths. It may also be poisoned by other things like webpack's operations. It could even be wrong due to something like the developer moving some files around. So, if we don't find the module using the whole module name, we should probably then try stripping the path and finding it using just the filename.

Anyway, I've got to go mow grass. I'll get back to this this evening (EDT). When I do so, my tentative plan is to think incremental - back off to the partial solution and write some issues.

trotzig:
I think that would be a good approach. I'd even be happy discontinuing support for rewrite-imports until we can find time to make it stable again.

trotzig:
@lencioni: do you know what to do about the codeclimate error?

trotzig:
@rhettlivingston: will you post a "I'm done, let's merge" comment when you are done pushing commits? Or just merge if you feel confident. :)

rhettlivingston:
Sure. Seems like I remember there are codeclimate fixes in the master that
aren't in our branch? If so, maybe we should merge the master to the
branch, make sure we're good, then merge to master.

trotzig:
Ah, you're right. I can do that before merging, unless you want to take it on.

rhettlivingston:
I'll give it a shot.

lencioni:
@trotzig I'm not sure why it errored. I'd say ignore it. I might remove that integration anyway, since it only runs eslint and we can't run eslint because they don't support all of the plugins we use yet. The code coverage should still work for the master branch.

lencioni:
@rhettlivingston I think your best bet would be to rebase this branch onto origin/master, not merge. But I think in this case you don't actually need to do any of that--you should be good to just click the merge pull request button (unless you end up pushing a new commit that has conflicts, but that seems unlikely).

rhettlivingston:
@lencioni Thanks. I sometimes err to the overabundance of caution side.

lib/ImportStatement.js
```diff
@@ -96,6 +102,9 @@ export default class ImportStatement {
 
     const importFunction = match.importFunction;
 
+    // Assume that an import with no assignment has side-effects.
+    const hasSideEffects = !match.assignment;
```
lencioni: Any import might have side-effects, so I'm a little worried about this naming. What do you think about changing it to something like `forSideEffectsOnly`?

lib/ImportStatements.js
```diff
@@ -9,13 +9,15 @@ import uniqBy from 'lodash.uniqby';
 import Configuration from './Configuration';
 import ImportStatement from './ImportStatement';
 
+const STYLE_SIDE_EFFECT = 'side-effect';
```
lencioni: Same comment about naming here. What do you think about `STYLE_SIDE_EFFECT_ONLY`/`side-effect-only`?

lib/ImportStatements.js
```diff
@@ -9,13 +9,15 @@ import uniqBy from 'lodash.uniqby';
 import Configuration from './Configuration';
```
lencioni: One minor nit: if it isn't too much trouble, could you configure your editor to wrap git commit messages at 72 characters?

lib/Importer.js
```diff
@@ -452,4 +435,25 @@ export default class Importer {
     }
     return messageParts.join(' ');
   }
+
+  injectSideEffectImports(importStatements: ImportStatements): Array<string> {
+    const addedImports = [];
```
lencioni: @trotzig why use an array here just to convert it back to a Set?

lib/ImportStatement.js
```diff
@@ -301,7 +303,7 @@ export default class ImportStatement {
     }
   }
 
-  _sideEffectImportString(): string {
+  _assignmentLessImportString(): string {
```
lencioni: The naming of this method is ambiguous to me now. How about `_importStringWithoutAssignment`?

package.json
```diff
@@ -66,6 +66,7 @@
     "eslint": "^2.8.0",
     "fb-watchman": "^1.9.0",
     "glob": "^7.0.3",
+    "lodash": "4.13.1",
```
lencioni: Unrelated: I was recently listening to a podcast with the maintainer of lodash and he was saying that he might deprecate the individual packages in favor of one big package. Plus the one big package receives more frequent updates than the individuals and has less duplication (since it doesn't simply inline dependencies). I think we should probably switch to the main package at some point.

lib/Importer.js
```diff
@@ -265,6 +272,11 @@ export default class Importer {
 
           newImports.push(jsModule.toImportStatement(this.config));
         });
+        // The "push" method on the ImportStatements class only accepts one ImportStatement
+        // at a time. Hence the need to loop.
```
lencioni: We should change the push method to accept multiple `ImportStatement`s at a time. I'm happy to add a commit to this PR that does that if you like.

lib/__tests__/Importer-test.js
```diff
@@ -3008,6 +3008,50 @@ bar
...
+
+bar
+          `.trim();
+        });
+
+        it('is passed through unchanged and sorted to the top', () => {
+          expect(subject()).toEqual(`
+import 'bar.html';
+
+bar
```
lencioni: I might be missing something here, but I'm not sure what this example is trying to test. Looking at the before and after, they seem to be identical.

trotzig: I think that's what we need to test. The `sorted to the top` is a bit confusing since we only have one import though.

lib/environments/meteorEnvironment.js
```diff
@@ -308,6 +308,31 @@ export default {
     return moduleName;
   },
 
+  moduleSideEffectImports: ({ pathToCurrentFile, config }: Object): Array<string> => {
+    let basePath = '';
+    if (pathToCurrentFile.endsWith('.js')) {
+      basePath = pathToCurrentFile.slice(0, -3);
+    } else if (pathToCurrentFile.endsWith('.jsx')) {
+      basePath = pathToCurrentFile.slice(0, -4);
```
lencioni: We have a `stripFileExtensions` configuration option that should be used here instead of these hard-coded extensions.

lencioni: Although since this is for the meteor environment, perhaps this doesn't apply? I don't have much context there, so I'll trust your judgement.

lib/Importer.js
trotzig: If I remember correctly, it was because I couldn't find a good way to merge two sets without first turning at least one of them into an array.

lib/environments/meteorEnvironment.js
```diff
@@ -308,6 +308,31 @@ export default {
...
+    } else {
+      return [];
+    }
+    const moduleSpecifiers = [];
+    for (const ext of ['.html', '.css']) {
```
lencioni: This would be better as a `.forEach` loop instead:

```js
['.html', '.css'].forEach(ext => {
  ...
});
```

or maybe a map and filter.

lib/__tests__/Importer-test.js
lencioni: Yeah, I think perhaps the wording should be adjusted.

lib/Importer.js
lencioni: 4f269e0

lib/ImportStatement.js
rhettlivingston: I was in-between on this. There are two possible indications of an import being included for its side-effects at this point - an existing side-effect only type of import or a specified environment included it in moduleSideEffectImports. If either of those has occurred, your earlier discussions convinced me that it should be sorted to the top because it has side-effects of an unknown nature.

I then realized that there is no technical reason why an import that has a side-effect and should be sorted to the top might not also resolve an unknown variable. This left me in the boat of having an import that is both known to have side effects and exports. Part of me says that the "hasSideEffrects" should stay true even though it is not "forSideEffectsOnly" because of your  earlier arguments that it would be safest to sort it to the top. Another part says to just forget about the side effects,,, but then I think that later code could choose to resolve that unknown via other modules that export it???

lib/ImportStatements.js
rhettlivingston: Same thought as above. If the environment tells me it has side-effects, should I sort it to the top regardless of whether it also has exports or not?

lib/ImportStatements.js
rhettlivingston: Absolutely. On a related side-note, do you want comments in code wrapped tighter than what I've been doing? It seemed that a lot of the existing comments are wrapped tighter, but ESLint is using 100 characters for the line-length. I personally got used to 132 back in the dark ages of print but accommodate whatever.

trotzig: I'm about to change to 100 chars (from previously 80 I think), are we okay with that?

rhettlivingston: I'm fine with any setting, just think that in code, our editors and ESLint
should match.

lencioni: I'm okay with 100 I guess--that's what we use at work. I've been resisting because I usually work in a terminal split where 80 characters is just right, but perhaps I just need to get with the times...

rhettlivingston: :D I understand. I can't focus on the laptop screen without reading glasses
anymore (who'd have thought that 51 is old) and have used multiple screens
since the 90s. I find that the modern 32" flat screen HDTVs are awesome
mounted on the wall at about 4 ft - cheapest and best monitors I ever
bought. And a side-by-side view of two files in atom fits perfect at about
100 characters though I have learned to habitually toggle the directory
view in and out. :D

package.json
rhettlivingston: That's interesting. I've recently had a discussion about a very similar package organization in material-ui and expressed my opinion that it just shouldn't be done because npm doesn't truly support it. They support one main in a package. And the "scoped package" feature falls short in not supporting a central initialization for the whole scope.

lib/__tests__/Importer-test.js
rhettlivingston: Copy and paste error. Will strip "sorted to the top".

lib/environments/meteorEnvironment.js
rhettlivingston: I felt that this was a different purpose??? "stripFileExtensions" appears to me to be for output formatting purposes only. The README has the note "Set to an empty array [] to avoid stripping out extensions." This would lead people to utilize that option who just like to see the ".js" and ".jsx" extensions explicitly stated in their import statements instead of assumed. If the user happened to be one of those folks and I was utilized stripFileExtensions in my search for side-effect imports, it would lead to looking for "someModule.js.html" instead of "someModule.html", thus breaking the above code. Reasonable?

lib/ImportStatements.js
rhettlivingston: Wow. Didn't imagine that smartgit wouldn't allow me to set commit message wrap points. I'll endeavor to write the messages elsewhere and paste them.

lib/environments/meteorEnvironment.js
rhettlivingston: Done.

rhettlivingston:
Started to add a test for the change in meteorEnvironments and stumbled on the first step, opening a file in my editor. Is anyone else starting to feel that a meteorEnvironment-test.js is needed? Or should a test for meteor's implementation of  moduleSideEffectImports appear in importer-test.js perhaps after the general test for that function in a more integrated than unit testing sort of way?

trotzig:
I think the Importer tests I added in 127700a6 and 9122d784 are sufficient for the integration part. A unit test for the logic added in MeteorEnvironment is probably enough now.

rhettlivingston:
The test addition was not insignificant. I'll give a bit of time for review before merge.

lib/__tests__/Configuration-test.js
```diff
@@ -409,6 +409,85 @@ describe('Configuration', () => {
     });
   });
 
+  describe('#get moduleSideEffectImports', () => {
+    describe('in meteor environment', () => {
+      describe('with no side-effect imports for a component', () => {
+        beforeEach(() => {
+          FileUtils.__setFile(path.join(process.cwd(), '.importjs.json'), {
+            useRelativePaths: true,
```
trotzig: This is irrelevant to the test, but no big deal.

trotzig:
The test looks good to me. I think this is ready to go.

lib/ImportStatements.js
lencioni: I've been thinking about editor configs and character widths. I think I'd like to keep my editor configured at 80 characters and then use the extra 20 characters that the linter rule allows as buffer space for troublesome sections of code. This allows for comments, for example, to follow a more readable line length for English but allows you to push the boundary for code when it is appropriate. Thoughts?

rhettlivingston: I can buy that. Does ESLint have a commentLineLength? Seems like we wouldn't be the only ones considering a split standard.

lencioni: It looks like the rule can take a comments option. http://eslint.org/docs/rules/max-len

rhettlivingston: Ask and you shall receive (should have known and looked though). If @trotzig agrees, I can buy moving that way.

Note that changing ESLint right now to have 80 char comments and 100 char code limits would probably reject a lot of comments I've made. If we all agree to go that way, I need to make some changes before it is turned on.

lencioni: I'm fine with 80 character comments and 100 character code limits. I'm not sure if it really matters much to adjust the linter for this, but I'd support that if you want to make that change.

trotzig: I'm okay with that change too.

PR #328, opened 2016-07-30
Fix bug in normalizePath, fix flow annotations

lencioni:
I found a small bug while debugging some things.

I also added some missing flow pragmas and adjusted some annotations. I tried to do this for lib/daemon.js as well, but that was causing vim to take 100% CPU mysteriously, so I avoided that for now.

### DIFF
--- a/lib/normalizePath.js
+++ b/lib/normalizePath.js
@@ -1,3 +1,5 @@
+// @flow
+
 export default function normalizePath(
   rawPath: string,
   workingDirectory: string
--- a/lib/normalizePath.js
+++ b/lib/normalizePath.js
@@ -5,11 +5,22 @@ export default function normalizePath(
   if (!rawPath) {
     return './';
   }
+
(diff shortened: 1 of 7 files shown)


lib/normalizePath.js
```diff
@@ -1,15 +1,28 @@
...
   if (!rawPath) {
     return './';
   }
+
   if (rawPath.startsWith(workingDirectory)) {
     return `.${rawPath.slice(workingDirectory.length)}`;
   }
-  if (rawPath.startsWith('.')) {
+
+  if (rawPath.startsWith('/')) {
```
lencioni: I'm not entirely sure if this is what we are supposed to do here, but it matches the previous behavior so I went with it. I think I tried out a version where it passed `rawPath` straight through if it starts with `/`, but that broke something. But I don't recall anymore.

trotzig:
Thanks @lencioni!

Issue #327, opened 2016-07-30
stripFromPath configuration modifies paths too early


lencioni:
I've noticed a weird symptom where my custom configuration only sometimes applies to my imports. By adding some logging in my .importjs.js file, I noticed that sometimes the path to the imported module would be like `'./path/to/foo.js'` while other times it is `'path/to/foo.js'`.

In this configuration, I happen to have a `stripFromPath` configuration that removes part of the path under certain conditions. I noticed that the `./` disappears after this is run. This leads me to believe that we are handling this at the wrong time.

For the purposes of configuration, paths should always be the same. I haven't had time to dig into this, further, but I think it probably makes sense to strip from path somewhere else. I think we need to take a more holistic look at how all of these parts fit together, come up with a good plan, and straighten it out.

lencioni:
I think we should remove `stripFromPath` and `stripFileExtensions` options in favor of `moduleNameFormatter`.

trotzig:
It's a little tricky to remove `stripFileExtensions` since people probably want that as a default. If they implement `moduleNameFormatter` without taking that into account, there might be confusion. I see that you didn't touch that in your work above, did you come to the same conclusion?

lencioni:
We have removed this option in favor of `moduleNameFormatter`, so I think I'm going to close this issue.

PR #329, opened 2016-07-31
Remove stripFromPath, lookupPaths, and useRelativePaths in favor of moduleNameFormatter

lencioni:
The `stripFromPath` option is mutating paths in between configuration
calls. This is no good because it means that configuration cannot rely
on paths being consistent when performing checks (e.g. when trying to
determine if something is a test file).

I originally started working to move this so it doesn't mutate the path
in JsModule at all. After I had that working, I realized that we
recently added a new configuration option that could be used to solve
the same use-case as this one: `moduleNameFormatter`. Since this is
currently causing bugs and since we already have a replacement ready, I
decided to simply remove this. This is a breaking change.

After I removed this, I realized that we have a few more options that
mutate the paths between configuration calls (`lookupPaths` and 
`useRelativePaths`), and which could be replaced by
`moduleNameFormatter` so I removed those as well.

If we decide to go forward with this, we should probably provide some
good examples of how to configure things to produce the desired effects.

We should add some tests that verify that paths don't change between
configuration checks, regardless of what they return.

We might also want to consider removing `stripFileExtensions` in favor of
a good default `moduleNameFormatter` function that does the same type
of thing.

Addresses #327

### DIFF
--- a/lib/environments/meteorEnvironment.js
+++ b/lib/environments/meteorEnvironment.js
@@ -359,13 +359,7 @@ export default {
     ['.html', '.css'].forEach((ext: string) => {
       const moduleSpecifier = `${basePath}${ext}`;
       if (fs.existsSync(path.join(config.workingDirectory, moduleSpecifier))) {
-        if (config.get('useRelativePaths')) {
-          moduleSpecifiers.push('.'.concat(path.sep, path.basename(moduleSpecifier)));
-        } else {
-          // Strip the leading '.' off of the moduleSpecifier to turn it into a
-          // Meteor compliant absolute path.
-          moduleSpecifiers.push(moduleSpecifier.slice(1));
-        }
+        moduleSpecifiers.push('.'.concat(path.sep, path.basename(moduleSpecifier)));
       }
     });
 
--- a/lib/environments/meteorEnvironment.js
+++ b/lib/environments/meteorEnvironment.js
@@ -359,7 +359,7 @@ export default {
     ['.html', '.css'].forEach((ext: string) => {
       const moduleSpecifier = `${basePath}${ext}`;
       if (fs.existsSync(path.join(config.workingDirectory, moduleSpecifier))) {
-        moduleSpecifiers.push('.'.concat(path.sep, path.basename(moduleSpecifier)));
+        moduleSpecifiers.push(`.${path.sep}${path.basename(moduleSpecifier)}`);
       }
     });
 
--- a/README.md
+++ b/README.md
@@ -106,7 +106,6 @@ configuration file in the root folder of your project.
 
 The following configuration options can be used.
 
-- [`lookupPaths`](#lookuppaths)
 - [`excludes`](#excludes)
 - [`aliases`](#aliases)
 - [`environments`](#environments)
@@ -123,21 +122,6 @@ The following configuration options can be used.
 - [`moduleNameFormatter`](#modulenameformatter)
 - [`tab`](#tab)
 - [`logLevel`](#loglevel)
-
-### `lookupPaths`
-
-Configure where ImportJS should look to resolve imports. If you are using
-Webpack, these should match the `modulesDirectories` configuration. Example:
-
-```javascript
-lookupPaths: [
-  'app/assets/javascripts',
-  'react-components',
-]
-```
-
-*Tip:* Don't put `node_modules` here. ImportJS will find your Node dependencies
-through your `package.json` file.
 
 ### `excludes`
 
@@ -239,8 +223,7 @@ memoize(() => { foo() });
 
 The key used to describe the named exports should be a valid import path. This
 can be e.g. the name of a package found under `node_modules`, a path to a
-module you created yourself without one of the `lookupPaths` prefixes, or a
-relative import path.
+module you created yourself, or a relative import path.
 
 ### `declarationKeyword`
 
@@ -326,8 +309,7 @@ import Foo from './foo';
 import Bar from '../baz/bar';
 ```
 
-Only imports located in the same `lookupPaths` will be made relative to each
-other. Package dependencies (located in `node_modules`) will not be imported
+Package dependencies (located in `node_modules`) will not be imported
 relatively.
 
 You can disable this by setting it to false:
--- a/README.md
+++ b/README.md
@@ -115,7 +115,6 @@ The following configuration options can be used.
 - [`groupImports`](#groupimports)
 - [`importDevDependencies`](#importdevdependencies)
 - [`importFunction`](#importfunction)
-- [`stripFromPath`](#stripfrompath)
 - [`stripFileExtensions`](#stripfileextensions)
 - [`useRelativePaths`](#userelativepaths)
 - [`ignorePackagePrefixes`](#ignorepackageprefixes)
@@ -307,17 +306,6 @@ importing](http://wiki.commonjs.org/wiki/Modules/1.1).
 importFunction: 'myCustomRequireFunction'
 ```
 
-### `stripFromPath`
-
-This option is used to trim imports by removing a slice of the path. The main
-rationale for using this option is if you have a custom `importFunction` that
-has different logic than the default `require` and `import from` behavior.
-
-```javascript
-stripFromPath: 'app/assets/',
-importFunction: 'requireFromAppAssets',
-```
-
 ### `stripFileExtensions`
 
 An array that controls what file extensions are stripped out from the resulting
@@ -481,7 +469,6 @@ When using `appliesFrom` only a subset of configurations are supported:
 - `declarationKeyword`
 - `importFunction`
 - `stripFileExtensions`
-- `stripFromPath`
 - `useRelativePaths`
 
 ## Dynamic configuration
--- a/README.md
+++ b/README.md
@@ -115,7 +115,6 @@ The following configuration options can be used.
 - [`importDevDependencies`](#importdevdependencies)
 - [`importFunction`](#importfunction)
 - [`stripFileExtensions`](#stripfileextensions)
-- [`useRelativePaths`](#userelativepaths)
 - [`ignorePackagePrefixes`](#ignorepackageprefixes)
 - [`minimumVersion`](#minimumversion)
 - [`maxLineLength`](#maxlinelength)
@@ -299,25 +298,6 @@ an empty array `[]` to avoid stripping out extensions.
 stripFileExtensions: ['.web.js', '.js']
 ```
 
-### `useRelativePaths`
-
-This option is enabled by default. When enabled, imports will be resolved
-relative to the current file being edited.
-
-```javascript
-import Foo from './foo';
-import Bar from '../baz/bar';
-```
-
-Package dependencies (located in `node_modules`) will not be imported
-relatively.
-
-You can disable this by setting it to false:
-
-```javascript
-useRelativePaths: false
-```
-
 ### `ignorePackagePrefixes`
 
 If you have package dependencies specified in `package.json` that are prefixed
@@ -417,7 +397,6 @@ through the `appliesTo` and `appliesFrom` options.
   {
     appliesTo: 'app/**',
     declarationKeyword: 'import',
-    useRelativePaths: true,
   },
   {
     appliesTo: 'app/**',
@@ -428,7 +407,6 @@ through the `appliesTo` and `appliesFrom` options.
     appliesTo: 'app/**',
     declarationKeyword: 'var',
     importFunction: 'mockRequire',
-    useRelativePaths: false,
   },
 ]
 ```
@@ -451,7 +429,6 @@ When using `appliesFrom` only a subset of configurations are supported:
 - `declarationKeyword`
 - `importFunction`
 - `stripFileExtensions`
-- `useRelativePaths`
 
 ## Dynamic configuration
 
@@ -481,23 +458,6 @@ module.exports {
       return 'const';
     }
     return 'import';
-  },
-}
-```
-
-Here's a more elaborate example taking both `pathToImportedModule` and
-`pathToCurrentFile` into account:
-
-```javascript
-module.exports {
-  useRelativePaths({ pathToImportedModule, pathToCurrentFile }) {
-    if (pathToCurrentFile.endsWith('-mock.js')) {
-      return false;
-    }
-    if (pathToImportedModule.endsWith('-test.js')) {
-      return false;
-    }
-    return true;
   },
 }
 ```
--- a/lib/Configuration.js
+++ b/lib/Configuration.js
@@ -35,7 +35,6 @@ const DEFAULT_CONFIG = {
   moduleSideEffectImports: (): Array<string> => [],
   stripFileExtensions: ['.js', '.jsx'],
   tab: '  ',
-  useRelativePaths: true,
   packageDependencies: ({ config }: Object): Array<string> =>
     findPackageDependencies(
       config.workingDirectory,
@@ -64,7 +63,6 @@ const RENAMED_CONFIGURATION_OPTIONS = {
   max_line_length: 'maxLineLength',
   minimum_version: 'minimumVersion',
   strip_file_extensions: 'stripFileExtensions',
-  use_relative_paths: 'useRelativePaths',
 };
 
 const KNOWN_CONFIGURATION_OPTIONS = [
@@ -87,7 +85,6 @@ const KNOWN_CONFIGURATION_OPTIONS = [
   'namedExports',
   'stripFileExtensions',
   'tab',
-  'useRelativePaths',
 ];
 
 const ENVIRONMENTS = {
@@ -338,11 +335,15 @@ export default class Configuration {
       hasNamedExports: true,
       variableName,
     });
-    if (this.get('useRelativePaths', { pathToImportedModule: relativeFilePath }) &&
-      !relativeFilePath.startsWith('meteor/') &&
-      !relativeFilePath.startsWith('node_modules')) {
-      jsModule.makeRelativeTo(this.pathToCurrentFile);
+
+    if (
+      relativeFilePath.startsWith('meteor/')
+      || relativeFilePath.startsWith('node_modules/')
+    ) {
+      return jsModule;
     }
+
+    jsModule.makeRelativeTo(this.pathToCurrentFile);
     return jsModule;
   }
 }
--- a/lib/Configuration.js
+++ b/lib/Configuration.js
@@ -35,7 +35,6 @@ const DEFAULT_CONFIG = {
   moduleNameFormatter: ({ moduleName }: Object): string => moduleName,
   moduleSideEffectImports: (): Array<string> => [],
   stripFileExtensions: ['.js', '.jsx'],
-  stripFromPath: null,
   tab: '  ',
   useRelativePaths: true,
   packageDependencies: ({ config }: Object): Array<string> =>
@@ -67,7 +66,6 @@ const RENAMED_CONFIGURATION_OPTIONS = {
   max_line_length: 'maxLineLength',
   minimum_version: 'minimumVersion',
   strip_file_extensions: 'stripFileExtensions',
-  strip_from_path: 'stripFromPath',
   use_relative_paths: 'useRelativePaths',
 };
 
@@ -91,7 +89,6 @@ const KNOWN_CONFIGURATION_OPTIONS = [
   'moduleSideEffectImports',
   'namedExports',
   'stripFileExtensions',
-  'stripFromPath',
   'tab',
   'useRelativePaths',
 ];
--- a/lib/Configuration.js
+++ b/lib/Configuration.js
@@ -29,7 +29,6 @@ const DEFAULT_CONFIG = {
   importDevDependencies: false,
   importFunction: 'require',
   logLevel: 'info',
-  lookupPaths: ['.'],
   maxLineLength: 80,
   minimumVersion: '0.0.0',
   moduleNameFormatter: ({ moduleName }: Object): string => moduleName,
@@ -62,7 +61,6 @@ const RENAMED_CONFIGURATION_OPTIONS = {
   ignore_package_prefixes: 'ignorePackagePrefixes',
   import_dev_dependencies: 'importDevDependencies',
   import_function: 'importFunction',
-  lookup_paths: 'lookupPaths',
   max_line_length: 'maxLineLength',
   minimum_version: 'minimumVersion',
   strip_file_extensions: 'stripFileExtensions',
@@ -82,7 +80,6 @@ const KNOWN_CONFIGURATION_OPTIONS = [
   'importDevDependencies',
   'importFunction',
   'logLevel',
-  'lookupPaths',
   'maxLineLength',
   'minimumVersion',
   'moduleNameFormatter',
@@ -344,7 +341,6 @@ export default class Configuration {
     if (this.get('useRelativePaths', { pathToImportedModule: relativeFilePath }) &&
       !relativeFilePath.startsWith('meteor/') &&
       !relativeFilePath.startsWith('node_modules')) {
-      jsModule.lookupPath = '.'; // hacky way to force relative paths
       jsModule.makeRelativeTo(this.pathToCurrentFile);
     }
     return jsModule;
--- a/lib/JsModule.js
+++ b/lib/JsModule.js
@@ -7,13 +7,6 @@ import ImportStatement from './ImportStatement';
 import requireResolve from './requireResolve';
 import resolveImportPathAndMain from './resolveImportPathAndMain';
 
-function normalizePath(pathToNormalize: string): string {
-  if (!pathToNormalize) {
-    return '';
-  }
-  return pathToNormalize.replace(/^\.\/?/, '');
-}
-
 // TODO figure out a more holistic solution than stripping node_modules
 function stripNodeModules(path: string): string {
   if (path.startsWith('node_modules/')) {
@@ -60,7 +53,7 @@ export default class JsModule {
     workingDirectory: string
   } = {}): ?JsModule {
     const jsModule = new JsModule();
-    jsModule.filePath = normalizePath(relativeFilePath);
+    jsModule.filePath = relativeFilePath;
 
     const importPathAndMainFile = resolveImportPathAndMain(
       jsModule.filePath, stripFileExtensions, workingDirectory);
--- a/lib/JsModule.js
+++ b/lib/JsModule.js
@@ -36,7 +36,6 @@ export default class JsModule {
    *   the project root.
    * @param {Array} opts.stripFileExtensions a list of file extensions to strip,
    *   e.g. ['.js', '.jsx']
-   * @param {String} opts.stripFromPath
    * @param {String} opts.variableName
    * @param {String} opts.workingDirectory
    * @return {JsModule}
@@ -47,7 +46,6 @@ export default class JsModule {
     makeRelativeTo,
     relativeFilePath,
     stripFileExtensions,
-    stripFromPath,
     variableName,
     workingDirectory = process.cwd(),
   }: {
@@ -56,7 +54,6 @@ export default class JsModule {
     makeRelativeTo?: ?string,
     relativeFilePath: string,
     stripFileExtensions: Array<string>,
-    stripFromPath?: string,
     variableName: string,
     workingDirectory: string
   } = {}): ?JsModule {
@@ -86,8 +83,6 @@ export default class JsModule {
     jsModule.variableName = variableName;
     if (makeRelativeTo) {
       jsModule.makeRelativeTo(makeRelativeTo);
-    } else if (stripFromPath) {
-      jsModule.stripFromPath(stripFromPath);
     }
     return jsModule;
   }
@@ -131,14 +126,6 @@ export default class JsModule {
     }
 
     this.importPath = importPath;
-  }
-
-  stripFromPath(prefix: string) {
-    if (!prefix) {
-      return;
-    }
-    this.importPath = this.importPath.replace(
-      RegExp(`^${escapeRegExp(prefix)}`), '');
   }
 
   displayName(): string {
--- a/lib/JsModule.js
+++ b/lib/JsModule.js
@@ -2,8 +2,6 @@
 
 import path from 'path';
 
-import escapeRegExp from 'lodash.escaperegexp';
-
 import Configuration from './Configuration';
 import ImportStatement from './ImportStatement';
 import requireResolve from './requireResolve';
@@ -16,19 +14,25 @@ function normalizePath(pathToNormalize: string): string {
   return pathToNormalize.replace(/^\.\/?/, '');
 }
 
+// TODO figure out a more holistic solution than stripping node_modules
+function stripNodeModules(path: string): string {
+  if (path.startsWith('node_modules/')) {
+    return path.slice(13);
+  }
+
+  return path;
+}
+
 // Class that represents a js module found in the file system
 export default class JsModule {
   hasNamedExports: ?boolean;
   importPath: string;
-  lookupPath: string;
   filePath: string;
   mainFile: ?string;
   variableName: string;
   workingDirectory: string;
 
   /**
-   * @param {String} opts.lookupPath the lookup path in which this module was
-   *   found
    * @param {Boolean} hasNamedExports
    * @param {String|null} opts.makeRelativeTo a path to a different file which
    *   the resulting import path should be relative to.
@@ -42,7 +46,6 @@ export default class JsModule {
    */
   static construct({
     hasNamedExports,
-    lookupPath,
     makeRelativeTo,
     relativeFilePath,
     stripFileExtensions,
@@ -50,7 +53,6 @@ export default class JsModule {
     workingDirectory = process.cwd(),
   }: {
     hasNamedExports?: boolean,
-    lookupPath: string,
     makeRelativeTo?: ?string,
     relativeFilePath: string,
     stripFileExtensions: Array<string>,
@@ -58,12 +60,11 @@ export default class JsModule {
     workingDirectory: string
   } = {}): ?JsModule {
     const jsModule = new JsModule();
-    jsModule.lookupPath = normalizePath(lookupPath);
     jsModule.filePath = normalizePath(relativeFilePath);
 
     const importPathAndMainFile = resolveImportPathAndMain(
       jsModule.filePath, stripFileExtensions, workingDirectory);
-    let importPath = importPathAndMainFile[0];
+    const importPath = importPathAndMainFile[0];
     const mainFile = importPathAndMainFile[1];
 
     if (!importPath) {
@@ -73,9 +74,6 @@ export default class JsModule {
     if (mainFile) {
       jsModule.filePath = path.normalize(`${importPath}/${mainFile}`);
     }
-
-    importPath = importPath.replace(
-      RegExp(`^${escapeRegExp(jsModule.lookupPath)}\/`), '');
 
     jsModule.importPath = importPath;
     jsModule.mainFile = mainFile;
@@ -102,21 +100,8 @@ export default class JsModule {
   }
 
   makeRelativeTo(makeRelativeToPath: string) {
-    if (this.lookupPath === undefined || this.lookupPath === null) {
-      return;
-    }
-
-    // Ignore if the file to relate to is part of a different lookupPath
-    if (!makeRelativeToPath.startsWith(this.lookupPath)) {
-      return;
-    }
-
-    // Strip out the lookupPath
-    const makeRelativeTo = makeRelativeToPath.replace(
-      RegExp(`^${escapeRegExp(this.lookupPath)}/`), '');
-
     let importPath = path.relative(
-      path.dirname(makeRelativeTo),
+      path.dirname(makeRelativeToPath),
       this.importPath
     );
 
@@ -129,7 +114,8 @@ export default class JsModule {
   }
 
   displayName(): string {
-    const parts = [this.importPath];
+    // TODO figure out a more holistic solution than stripping node_modules
+    const parts = [stripNodeModules(this.importPath)];
     if (this.mainFile) {
       parts.push(` (main: ${this.mainFile})`);
     }
@@ -169,8 +155,10 @@ export default class JsModule {
       defaultImport = this.variableName;
     }
 
-    const pathToImportedModule = this.resolvedFilePath(config.pathToCurrentFile,
-                                                       config.workingDirectory);
+    // TODO figure out a more holistic solution than stripping node_modules
+    const pathToImportedModule = stripNodeModules(this.resolvedFilePath(
+      config.pathToCurrentFile, config.workingDirectory
+    ));
 
     return new ImportStatement({
       declarationKeyword:
@@ -181,7 +169,8 @@ export default class JsModule {
       namedImports,
       path: config.get('moduleNameFormatter', {
         pathToImportedModule,
-        moduleName: this.importPath,
+        // TODO figure out a more holistic solution than stripping node_modules
+        moduleName: stripNodeModules(this.importPath),
       }),
     });
   }
--- a/lib/__mocks__/findMatchingFiles.js
+++ b/lib/__mocks__/findMatchingFiles.js
@@ -2,7 +2,7 @@ import formattedToRegex from '../formattedToRegex';
 
 let existingFiles = [];
 
-export default function findMatchingFiles(lookupPath, variableName) {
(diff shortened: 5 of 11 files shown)


lencioni:
I am seeing that pathToCurrentFile stats with "./" and pathToImportedModule doesn't. Not sure if that's related to this change.

lencioni:
I believe that this problem comes from this line in JsModule:

```js
 jsModule.filePath = normalizePath(relativeFilePath);
```

`normalizePath` here simply strips the leading `./` or `.` from the path if it exists. I believe that this is in there in order to support the `lookupPaths` option--otherwise, we could just leave the `./` in there and every path would be relative to the project root here, which simplifies things.

Taking a step back, `lookupPaths` causes confusion, and it seems like the functionality could be solved with `moduleNameFormatter`. I would like to also remove this option.

I may be wrong, but I think that `useRelativePaths` can also be accomplished by `moduleNameFormatter`, so I'd also like to remove that option.

lencioni:
We might also be able to remove `stripFileExtensions` and just rely on the module formatter option.

In this case, it might be nice to provide a default module formatter that strips some extensions by default.

lencioni:
If we decide to make all of these breaking changes, we might as well go all in and remove the old `appliesTo` and `appliesFrom` options.

lib/environments/meteorEnvironment.js
```diff
@@ -321,13 +321,7 @@ export default {
     ['.html', '.css'].forEach((ext: string) => {
       const moduleSpecifier = `${basePath}${ext}`;
       if (fs.existsSync(path.join(config.workingDirectory, moduleSpecifier))) {
-        if (config.get('useRelativePaths')) {
-          moduleSpecifiers.push('.'.concat(path.sep, path.basename(moduleSpecifier)));
-        } else {
-          // Strip the leading '.' off of the moduleSpecifier to turn it into a Meteor compliant
-          // absolute path.
-          moduleSpecifiers.push(moduleSpecifier.slice(1));
```
lencioni: I'm really not sure about this particular change. @rhettlivingston can you weigh in on whether this is important?

rhettlivingston:
I have no love for stripFromPath, stripFileExtensions and lookupPaths and believe that a more generalized approach to providing their functionality would be great. But I'm a little thrown by the idea that moduleNameFormatter can replace them.

I divide the options up in my head into groups. One of those is a general control group into which I place     environments, minimumVersion, logLevel, appliesTo, and appliesFrom. The other two are a) options that effect resolution of a module versus b) those that effect output of the new file after resolution has occurred. moduleNameFormatter is at the end of the game, after resolution has occurred. It gives the user code (or environment code) an opportunity to control how its moduleNames are output in the import statements. We wouldn't want to utilize its output for anything other than printing. I thought that all three of the other options mentioned either expressed where to find files or how to mutate them so that a variable might match them. So they all have effect in the resolution process.

I do see how lookupPaths has an output effect. I would be fully in on the idea of an "includes" option that does nothing but limit our search to replace that aspect of lookupPaths, and allow its output formatting functionality (that of expressing the moduleName relative to the lookupPath instead of the project root) to be performed in moduleNameFormatter.

But if the stripXXX options can be replaced by moduleNameFormatter, I'm totally misunderstanding them.

lib/environments/meteorEnvironment.js
rhettlivingston: I'm a little iffy about removing useRelativePaths. I agree that it can be implemented within moduleNameFormatter. My problem with doing that is that it seems like something that is too common for a user to want to change. If they change it, they'll be burdened not just with implementing that in their user config, but also with carrying over anything else the environment implemented in moduleNameFormatter.

I'm probably coming from an unusual perspective though because npm packages likely almost always use relative. Meteor has started off relative, but there are some big groups leaning hard towards absolute.

When writing an app using ES6 modules, the need for local packages is gone, modularization can be accomplished within the app without packages. This then leads you to larger directory structures. Users are finding it more convenient to use absolute paths because when they decide to juggle that directory structure up, they can easily do a global search and replace on an absolute reference. That can't be done with relative references.

lencioni: > If they change it, they'll be burdened not just with implementing that in their user config, but also with carrying over anything else the environment implemented in moduleNameFormatter.

Perhaps we need to chain these. So instead of overriding the base `moduleNameFormatter` and  the environment `moduleNameFormatter` when you define your own, they all happen. Or, maybe they can make decisions about if the underlying formatter happens (e.g. passing in the next `moduleNameFormatter` functions as an argument, as @trotzig had suggested).

If we do that, it would be pretty easy to keep the option but to implement it in the base `moduleNameFormatter` if we wanted.

lencioni:
I think there might be a mismatch between what these options were intended for and how they may have ended up being used. It also seems as though we haven't implemented them very cleanly, so they end up having these effects on module resolution. As far as I know, that was unintentional.

`lookupPaths` was added to support generating module paths that would work well with webpack configurations where you can specify the module directories webpack looks in. This was added before import-js supported relative paths at all, so it was used to set the root of the paths that would be generated. I'm not entirely sure, but I believe that the fact that it affects file finding was done as an optimization. This may end up producing duplicate matches, where we take the shortest path, which is a bit awkward.

I believe that the intended use-case for this can be implemented in `moduleNameFormatter` by doing something like:

```js
moduleNameFormatter({ pathToImportedModule, moduleName }) {
  const lookupPath = './some/lookup/path/';
  if (pathToImportedModule.startsWith(lookupPath)) {
    return pathToImportedModule.slice(lookupPath.length);
  }

  return moduleName;
},
```

`stripFromPath` was another option that was implemented as a way to have more control over the module path. The use-case was a project where we had a custom helper require function that was used in test files to require app files instead of the built-in `require`. This custom helper was used to make it so paths didn't have to be relative (in this project, relative paths were undesirable because tests and app code were in different mirrored directory trees, so relative paths for app code included a lot of `../../../` type stuff). The configuration here, for example, makes it so when you import an app module (in app/assets/javascripts) from a spec (in spec/javascripts), it will use the custom helper function, which accepts a string path name that starts in the root of the JS directory (e.g. if the file was at app/assets/javascripts/foo.js, this sting should be "foo.js").

This can be accomplished with a `moduleNameFormatter` like:

```js
moduleNameFormatter(paths) {
  if (isTestImportingAppModule(paths)) {
    return paths.moduleName.replace(/^(..\/)*app\/assets\/javascripts\//, '');
  }
  return paths.moduleName;
},
```

`stripFileExtensions` is pretty straightforward to implement this way as well:

```js
moduleNameFormatter({ moduleName }) {
  return moduleName.replace(/\.jsx?$/, '');
},
```

And `useRelativePaths`:

```js
moduleNameFormatter({ pathToImportedModule }) {
  return pathToImportedModule.slice(2); // remove leading `./`
},
```

lencioni:
I'm going to cherry pick some of the smaller commits onto master and rebase this to make the PR a little easier to review.

lib/environments/meteorEnvironment.js
trotzig: What if we changed `moduleNameFormatter` into an array of `moduleNameTransformers`? 

```
moduleNameTransformers: [
  ({ moduleName }) => moduleName.replace(/\jsx?$/, ''), // what used to be stripFromPath
  ({ moduleName }) => moduleName.replace(/^app/, ''), // what used to be lookupPath
],

We could merge this with whatever is in the default configuration/environment. 

The only option I have a hard time seeing here is `useRelativePaths`, because that's something that we want by default, but allow people to turn off.

File: lib/environments/meteorEnvironment.js [hunk 2]
```diff
@@ -321,7 +321,7 @@ export default {
     ['.html', '.css'].forEach((ext: string) => {
       const moduleSpecifier = `${basePath}${ext}`;
       if (fs.existsSync(path.join(config.workingDirectory, moduleSpecifier))) {
-        moduleSpecifiers.push('.'.concat(path.sep, path.basename(moduleSpecifier)));
+        moduleSpecifiers.push(`.${path.sep}${path.basename(moduleSpecifier)}`);
```
trotzig: I think we can do `path.join('.', path.basename(moduleSpecifier))` as well.

lib/environments/meteorEnvironment.js
lencioni: To "turn off" `useRelativePaths` with this, you basically need to return a formatted version of `pathToModule`. Here's a simplified example:

```js
moduleNameFormatter({ pathToImportedModule }) {
  return pathToImportedModule.slice(2); // remove leading `./`
},
```

I think the array approach might work, but I don't like the special casing of one config value. I think it would be better if we can find a solution that will work for any configuration value. I have some thoughts about this.

It looks like we currently execute configurations starting with the user config, followed by the environment config, and ending with the base config. At each level of this configuration, we make a decision about whether we should merge the result or just return the value. This special cases some configuration options in potentially confusing ways. Also, it isn't as flexible as it could be. What if an environment configuration wanted to override some of the base config's values instead of simply being merged in? Or what if the user config wanted to do the same?

What if we reversed the order that configurations are executed (base, env, then user) and always pass down the previous value (similar to what we have done for `moduleName` and `moduleNameFormatter`, but more general, so the object would look more like `{ pathToImportedModule, pathToCurrentFile, value }`). This would allow environment and user configs to merge in previous values if they want, or to override them, which allows us to remove the special casing `MERGABLE_CONFIG_OPTIONS`.

Once this is done, we could choose to implement convenience configurations like `stripFileExtensions` in the base config's `moduleNameFormatter`. 

I think we should also remove the JSON configuration and require that every config value be a function, to simplify the logic in our `Configuration` class, but I don't think that is necessary for this.

lib/environments/meteorEnvironment.js
lencioni: Good call. That's better.

lencioni: Actually, that doesn't work. 

```js
> path.join('.', 'foo')
'foo'
```

trotzig: 😞

trotzig:
I've played around a little with these changes on the brigade codebase, and I'm seeing a few issues. First, here's my configuration:

```js
const WEBPACK_ROOTS = [
  './app/assets/javascripts/',
  './app/assets/',
  './spec/javascripts/'
];

function stripWebpackRoots(path) {
  for (let i in WEBPACK_ROOTS) {
    if (path.startsWith(WEBPACK_ROOTS[i])) {
      return path.slice(WEBPACK_ROOTS[i].length);
    }
  }
  return path;
}

function stripFileSuffix(path) {
  return path.replace(/\.jsx?$/, '');
}

module.exports = {
  moduleNameFormatter: ({ pathToImportedModule }) =>
    stripFileSuffix(stripWebpackRoots(pathToImportedModule)),
  // plus some more
};
```

It seems that files in `app/assets/components` aren't being stripped:

```js
const AlertsPage = require('components/pages/AlertsPage');
```

becomes 

```js
const AlertsPage = require('app/assets/components/pages/AlertsPage/index');
```

There's a second issue in there too, around the index.js. I guess I could strip that out via `moduleNameFormatter` although it feels a little clunky.

I'm also getting full paths to files if I have to pick from a list:
```js
const History = require('lib/history');

**Comment by trotzig** [comment]
Whoops, I hit the submit button too early. Here's a continuation:

```js
const History = require('lib/history');
```

becomes 

```js
const History =
  require('/Users/henrictrotzig/brigade/app/assets/javascripts/lib/history');
```

after picking in the list. 

The node_modules folder doesn't seem to be ignored:

```js
const React = require('react');
```

becomes 

```js
const React = require('react/react');
```

lib/environments/meteorEnvironment.js
rhettlivingston: I concur heartily with the change in order. I reviewed quite a few configuration implementations in npm about a month ago and they all went in that direction - for example [node-config's order documentation is here](https://github.com/lorenwest/node-config/wiki/Configuration-Files).

Most also supported many file formats and even allowed ENV and command-line overrides as the last level considered, but I understand the reasons that without a 3rd party package managing this we might want to fall back to just js.

I also concur with having some system in which a function at any level can choose to either inherit or supercede what was produced by the previous level. Many packages seem to have functionality like that.

rhettlivingston:
Just to throw in a thought I've had for a while (FWIW), it feels odd that stripFileExtensions isn't a boolean. If we throw it into moduleNameFormatter's domain, that doesn't matter, but the reason it feels odd does. It feels odd because using it to strip something like ".mjs" would be meaningless because ".mjs" isn't an accepted file extension.

That points out that we don't have an option for defining the file extensions to be considered during a search for modules. That seems basic. The filename extension to aliases is sort of in that direction, but wasn't really meant for that and doesn't really get us there. There is at least one point in code with a hard-coded *.js* which doesn't seem ideal. It also doesn't match some file-extensions I've seen, including the ".mjs" extension that the node guys have been discussing in some of the proposed node standards for identifying ES6 modules.

As in other recent discussions, this gains a little more importance with the addition of #199.

lencioni:
@trotzig 

> It seems that files in app/assets/components aren't being stripped

That's odd, and I bet it is a bug in our code. I think it is important for `pathToImportedModule` and `pathToCurrentFile` to be consistent, which means that they should be relative to the project root and start with `./`. It seems that for these paths for you, they do not start with `./` and we should get to the bottom of this. You can verify by adding a console.log in your module formatter function and tailing the importjsd logs.

> There's a second issue in there too, around the index.js. I guess I could strip that out via moduleNameFormatter although it feels a little clunky.

Yeah, I forgot about this. True, it is pretty easy to strip out, but it is one more thing to take into account. This regex would do it:

```js
/(?:\/index)?\.jsx?/
```

> I'm also getting full paths to files if I have to pick from a list

I think this is another bug that we need to fix anyway.

> The node_modules folder doesn't seem to be ignored

I don't think I quite follow. In your example you said you have the path `react/react`. Did you mean `node_modules/react`?

Is this your whole configuration?

I think that these are all solvable problems. What I'd like for us to decide is whether we think this is a step in the right direction or not.

@rhettlivingston 

> It feels odd because using it to strip something like ".mjs" would be meaningless because ".mjs" isn't an accepted file extension.

That does feel odd, but you can configure your environment (e.g. webpack or node) to not require any extensions you like. This generally includes ".js" but many people have set it up to not require ".jsx" as well. I imagine that folks might want to do this for any number of extensions. At Brigade, we were using ".js.jsx" for a while IIRC. I think that using a list is necessary.

> That points out that we don't have an option for defining the file extensions to be considered during a search for modules.

True, this is a current limitation. It might be interesting to expand this to be configurable, but I'd like to wait until we know it is actually in demand. If TC39 moves forward with an official .mjs extension, we can certainly add it without making this configurable.

rhettlivingston:
@lencioni Hmmm. I think I was misunderstood just a little.

I too think that a list is necessary. I just think that the list should be a list of accepted extensions like `extensions: ['.js', '.jsx', '.json', '.js.jsx', '.whatever']` and that stripFileExtensions can then be `stripFileExtensions : true` or `stripFileExtensions: false`. It doesn't make sense to be able to strip extensions that aren't recognized by our local module search code. I also think that you had the demand and just solved it with \*.js\* which matches .js, .jsx, .json, .js.jsx and something like `this.jshow.html`. 

On your side of the equation though I can also see the possibility that a project might want to have some extensions stripped and some not which your current syntax allows.

lencioni:
Yeah, that's important. Especially if you think about things like `.coffee`, which you would normally want to match but not strip.

rhettlivingston:
Interesting that you bring that up. I had forgotten that `.coffee` is a relatively common extension in Meteor world and that they've extended the syntax a little to add ES6 module support. Are we somehow handling coffee files now? They don't match our search string.

trotzig:
We don't at the moment.

trotzig:
I went ahead and moved the discussion around `stripFileExtensions` to #334.

trotzig:
> Yeah, I forgot about this. True, it is pretty easy to strip out, but it is one more thing to take into account. 

In addition to `index.js`, I think you'd have to take care of paths ending in `package.json` as well. Ninja edit: actually not, it looks like we strip that out in `lib/JsModule.js`.

> I don't think I quite follow. In your example you said you have the path react/react. Did you mean node_modules/react?

Nope, but I know now why it ended up as `react/react`. We send all imports through the `moduleNameFormatter` function, even package dependencies. This means that in order to get `moduleNameFormatter` right, I need to match the `moduleName` against package dependencies so that I can avoid using `pathToImportedModule`. 

> Is this your whole configuration?

Nope, I left out things that weren't significant. Here's the full file, as it currently needs to be in order to produce the same output as before:

```js
const WEBPACK_ROOTS = [
  './app/assets/javascripts/',
  './app/assets/',
  './spec/javascripts/',
  'app/assets/javascripts/',
  'app/assets/',
  'spec/javascripts/',
];

function stripAbsolutePath(path, workingDirectory) {
  if (path.startsWith(workingDirectory)) {
    return path.slice(workingDirectory.length + 1); // +1 for the slash
  }
  return path;
}

function stripWebpackRoots(path) {
  for (let i in WEBPACK_ROOTS) {
    if (path.startsWith(WEBPACK_ROOTS[i])) {
      return path.slice(WEBPACK_ROOTS[i].length);
    }
  }
  return path;
}

function stripFileSuffix(path) {
  return path.replace(/\.jsx?$/, '');
}

function stripIndexJs(path) {
  return path.replace(/\/index\.jsx?$/, '');
}

module.exports = {
  moduleNameFormatter: ({ pathToImportedModule, moduleName, config }) => {
    // Check if this is a package dependency, e.g. `react`, `jquery`. In such
    // case, we don't want to alter the module name.
    if (config.get('packageDependencies').includes(moduleName)) {
      return moduleName;
    }

    return stripFileSuffix(
      stripWebpackRoots(
        stripIndexJs(
          stripAbsolutePath(pathToImportedModule, config.workingDirectory))));
  },
  minimumVersion: '0.5.0',
  aliases: {
    $: 'jquery',
    branch: 'imports?define=>false!branch-sdk',
    Uri: 'urijs',
    styles: './index.scss',
    Route: 'react-router/lib/Route',
    IndexRoute: 'react-router/lib/IndexRoute',
    Redirect: 'react-router/lib/Redirect',
    qs: 'querystringify',
  },
  ignorePackagePrefixes: [
    'react-',
    'react-addons-',
  ],
  namedExports: {
    'constants/app_constants': [
      'ActionTypes',
    ],
    'react': [
      'PropTypes',
    ],
    'react-intl': [
      'FormattedDate',
      'FormattedMessage',
      'FormattedNumber',
      'FormattedRelative',
      'FormattedTime',
      'IntlProvider',
    ],
    'react-motion': [
      'spring',
    ],
    'underscore': [
      'clone',
      'compact',
      'debounce',
      'difference',
      'groupBy',
      'isEmpty',
      'isEqual',
      'isObject',
      'last',
      'mapObject',
      'max',
      'memoize',
      'object',
      'omit',
      'partial',
      'pick',
      'pluck',
      'range',
      'shuffle',
      'sortBy',
      'tap',
      'throttle',
      'times',
      'values',
      'without',
    ],
  },
  declarationKeyword: 'const',
};
```

It's a little awkward having to deal with package dependencies. I wonder if we could just not call `moduleNameFormatter` for those. I'm not sure. 

We need to fix the inconsistencies around `pathToCurrentModule` before we can move along with this PR. Other than that I think it's good, although my gut is telling me that this is perhaps a bit too aggressive right now.

lencioni:
> Nope, but I know now why it ended up as react/react. We send all imports through the moduleNameFormatter function, even package dependencies. This means that in order to get moduleNameFormatter right, I need to match the moduleName against package dependencies so that I can avoid using pathToImportedModule.

Ah, of course. As a heuristic, you could first see if `moduleName` starts with "./" or "../" instead of looking at dependencies. Or... what is `pathToImportedModule`? Is it something like "./node_modules/react/react.js" or is it more like "react/reactjs". You could use that as a heuristic as well.

Alternatively, we could pass more metadata about the file to the config functions.

rhettlivingston:
>Or... what is pathToImportedModule?

From meteorEnvironments moduleNameFormatter routine (strategy found somewhere else in the codebase)...
```js
    // If the module being imported is an npm package, the path to the module will start with
    // 'node_modules/' and the moduleName should not be altered.
    if (pathToImportedModule.startsWith('node_modules/')) {
      return moduleName;
    }
```

trotzig:
I've been working on fixes to the inconsistent `pathToCurrentModule`. I've only found the cause of the absolute path so far, which I believe is because of how `path.resolve` works in `JsModule#resolvedFilePath`. Basically, when we send down options for the user to choose from, we now use the relative paths. When that path comes back as selected, we have lost the path to the file. So we try to resolve it again, using `path.resolve`. I think we should be able to just normalize that path through `normalizePath`. Which I tried, but ran into some weird cycling leaving vim unresponsive. I'll continue looking at that when I have more time.

rhettlivingston:
>Basically, when we send down options for the user to choose from, we now use the relative paths. When that path comes back as selected, we have lost the path to the file. So we try to resolve it again, using path.resolve

That corresponds to a feeling that I had while trying to implement full support for side-effect imports in the rewrite code that we badly need to cleanly and clearly separate the true paths to files from the paths that are used in the output of import statements, i.e. filePath versus displayPath or... actually modulePath / moduleName does it for me because moduleName corresponds to the element name in the language specification.

Perhaps we already have that separation and have just misapplied it, but I suspected we're missing needed storage elements.

trotzig:
I agree about the need the clean separation. For new imports that we construct by finding a file in the file system, this is straightforward. The tricky part is figuring out file paths for existing imports, or imports that we have an alias for. Which I guess speaks to the current limitations of our aliases implementation. Perhaps they need to be paired with a full file path? In our case (brigade), we have an alias that can't be resolved to a file unless we bring in webpack to do the job: 

```json
    "branch": "imports?define=>false!branch-sdk",`
```

> Perhaps we already have that separation and have just misapplied it, but I suspected we're missing needed storage elements.

What do you mean by "storage elements" in this context? The separation is there, although it's not very clean. And it's buggy, as we've seen through exploring this PR. 

I like naming them `moduleName`/`modulePath`. We currently send down `moduleName`/`pathToImportedModule` to config functions, and we still have a chance to rename that before we release the next version. It should be reflected elsewhere in the code as well of course.

rhettlivingston:
Yeh, I know TL;DR. View this as an attempt to gather thoughts, some of which are fleeting and not fully formed. Take all with a grain of salt because I'm not checking facts super thoroughly and some have probably changed. :smile:

One note on the "bugginess". I don't think "useRelativePath" and other functions related to creating a new moduleName in an import statement that we are outputting have any business having effects within the JsModule class. That class should be for representing the facts about a module. That might include storing an importPath (moduleName) found in an import statement, but not creating a new one. How to refer to a module in an import statement seems to me to be the job of the ImportStatement class.

I think the issues that I felt while trying to deal with side-effect modules may be rooted deeper in JsModule and the infrastructure that works with it such as findJsModuleFor, findMatchingFiles, etc. So, I just want to explore that for a moment (or an eon).

At the highest level of my thoughts, I think there should be a container for JsModule's. And it should have searches (or a search that accepts multiple parameters or a function) that allow us to easily find all known JsModules matching a bindingIdentifier (aka namedExport and ExportName (another spec-based name)), a moduleName, and/or a modulePath. It should not assume that there is only one match for either. This pool could aid greatly in not repeating our scans of directories and files. It probably should contain much of the code that fills it also, including the functionality of findJsModulesFor and findMatchingFiles. Our finds should go through this pool first, possibly only. It should be fully built in our discovery phase before we start trying to build importStatements for output.

The JsModule's class description is "Class that represents a js module found in the file system". I don't think that description is correct nor should it be. IMO, it needs to be more like "class that represents a known or speculated js module". There are several sources for JsModules and we can't always find the resulting module in the system. Sources include at least namedExports and aliases from the configuration, existing import statements, hard-coded knowledge (core modules), package systems, and local modules.

Problem zero (had to insert one after the fact :smile:) - if it can be used in a moduleName field of an import statement, its a module and JsModule needs to be able to represent it and its interface definition.

Problem number one, JsModule doesn't do an adequate job of representing modules we know about but can't find (or possibly have found multiple possibilities for). This category includes webpack modules, core modules, sometimes packages, and I'm sure there are others. I think it should be able to hold all of those and also remember whether they are core, package, or local if that is known.

Problem number two - Side-effect modules. We usually know where they are, but they don't have variableNames and variableName is a required property of the class. It shouldn't be.

Problem number three, JsModule badly needs to be extended to contain everything we need to know about a module. It currently represents a module / variable pair, not a module. It needs a namedExports array. I believe we could get rid of the variableName and include nothing but a namedExports array. If there is a default export, simply add a "default" entry to the namedExports array. This precisely matches the ES6 standard's treatment of default.

Problem number four - it needs an aliases dictionary. Actually, extend the namedExports to be an array  of objects. Each object is an export and known aliases for that export (if any). That way, we can't accidentally create an alias for something not exported by the module. The current aliases implementation corresponds to the namedExport "default". We can already pick up other aliases from import statements. Note that we shouldn't store export aliases. They are a different animal. The export alias ends up being the namedExport and we forget about the internal identifier it aliased.

Possible problem - I wonder if the filePath (modulePath) should be truly absolute. If not, perhaps lookupPath should be. I have a case where I'm finding modules in a system level directory. npm also has a global source. These are packages of course, but, as I indicated above, I think we should be able to store the interfaces of all modules in this structure, not just local modules. If we've found a path to a module of any type, we need to be able to keep it and there is no reason it should ever change (or rather we should assume files don't mutate during a run).

Speculated problem - we mutate found facts instead of storing them as "original" or something and associating them with a source. At the least, this makes debugging tougher. I'm not going to include original fields in the straw-man below, but its own my mind.

Thought: it may be that "merging" of info from different sources should be performed in the JsModule constructor. It could scan JsModules to see if there is an existing match (the current info matches and has no conflicts with the known info in an existing JsModule), possibly merge info into that, and return it. If no existing match is found, then you return something new.

```js
JsModule : {
  // Naming and Info
  //-------------------
  // 'meteor/somepackage', 'fs', 'path', '../../otherdir/somelocalfile.js',
  // 'imports?define=>false!branch-sdk'etc. 
  // Note, I'm thinking we shouldn't store a speculative moduleName here when
  // building a JsModule from a file scan. There is nothing in the ES6
  // specification defining how a filename relates to a moduleName or even if
  // there is always a file. They intentionally left it undefined.
  // This is primarily from import statements and configuration information.
  // It is up to importStatement to choose a moduleName to output.
  moduleName,
  moduleCategory, // core, package, local (if known)
  // a list of sources that have contributed to this? Seems useful for 
  // debugging. It could include fileScan (created on basis of fileName alone),
  // fileExtraction (extracted an xface definition from it), aliases,
  // namedExports, importStatement, etc.)
  sources, 

  // location(s)
  //-------------------
  // I wonder if we should define a location object having all of these
  // properties and make this an array of location objects in some sort of
  // likely-true-first sort order.
  // Problem is, we'd then be in a position of having multiple JsModule objects
  // that may or may not be for the same location. That is a truth though. We'll
  // never be successful in merging every separate piece of information we get
  // that relates to the same real-world JsModule. So, we need to accept the
  // reality.
  // For example, we would certainly be able to create a JsModule for any file
  // we find. We may also have the same JsModule from an importStatement or
  // other source and not be able to definitively merge the two as happens with
  // webpack for example.
  // My usual policy is "when in doubt, store it" and use algorithms to see
  // through the doubt at usage.
  modulePath,
  mainFile, // wonder if this should be = modulePath for local modules
            // i.e. make it something that always points to a js file and let
            // modulePath point to either files or some package element.
  lookupPath, // ? is this still useful when outputting a module name in a new 
              // import statement or something ?)

  // interface definition
  //----------------------
  // Note: it might be worth it to have the records in this be "ExportEntry"s
  //    as defined in the ES7 spec at [table 42](https://tc39.github.io/ecma262/#table-41).
  //    The thing is, only 3 out of 4 of those fields have potential use here
  //    (the localName has no use to us). Of the others, if we choose to have
  //    the recursion for "import *" performed at extraction time, we don't
  //    need anything other than ExportName (which corresponds to what we
  //    currently keep in namedExports). If we chose to keep the relationship
  //    defined by 'import * from xxxxx', we'd need the other two fields.
  namedExports[    // note, could just as well be named "interface"
    {
      bindingIdentifier,
      // could be different ones from existing import statements and
      // configuration. Do we need to remember where each came from? They may
      // have differing reliabilities. An alias from an import statement for
      // example is very likely to be depended on by that module's naming and
      // would likely have the highest degree of "truthiness".
      // Note that as alias is actually not a true part of the interface. It is
      // applied in the import. So, they are importing-module-specific. I
      // I thought about the fact that that could mean they should be kept in
      // a structure outside JsModule. I'm of the opinion that that is being
      // overly precise in representation. But, if we ever use a JsModule to
      // create imports in multiple files as might happen if we created an ultra
      // fast fix-all capability, we'd need to remember which consuming file an
      // alias is locked to (if any, the ones from the config may be universal).
      aliases: [<alias>],
    },
    {
      bindingIdentifier,
      aliases: [<alias>],
    },
  ],
  usedForSideEffects, // true if we've ever seen an indicator that this module
                      // is used primarily for side-effects. i.e. side-effect 
                      // import or config.moduleSideEffectImports said so.
}
```

Here is a related strawman I sketched as a configuration option corresponding to the JsModule that could replace namedExports, aliases, and maybe some other things with something that is more capable and roughly standard-based.

```js
modules = {
  <moduleName>: {
    exportEntries: [ // trying out another standard-based name for namedExports
      {
        // The name used by the module to export the binding. Can be "default".
        exportName, 
        importAliases: [ // optional field for defining import-side aliases for exportNames
          {
            importLocalName,
            applicableTo: [] // optional globs or regexps describing what
                             // consumers use this alias. Default universal.
          }
        ],
      }
    ],
    modulePath: <absolutePath>,    // optional path to module
    usedForSideEffects: <boolean>, // optional boolean. Default false.
  },
  ...
}
```

applying that to the aliases and namedExports in an earlier example in this discussion...
```js
const modules = {
  'constants/app_constants': {
    exportEntries: [{ exportName: 'ActionTypes' }],
  },
  'imports?define=>false!branch-sdk': {
    exportEntries: [{ exportName: 'default', importAliases: ['branch'] }],
  },
  './index.scss': {
    exportEntries: [{ exportName: 'default', importAliases: ['styles'] }],
  },
  jquery: {
    exportEntries: [
      { exportName: '$' },
      { exportName: 'jquery' },
    ],
  },
  querystringify: {
    exportEntries: [{ exportName: 'default', importAliases: ['qs'] }],
  },
  react: {
    exportEntries: [{ exportName: 'PropTypes' }],
  },
  'react-intl': {
    exportEntries: [
      { exportName: 'FormattedDate' },
      { exportName: 'FormattedMessage' },
      { exportName: 'FormattedNumber' },
      { exportName: 'FormattedRelative' },
      { exportName: 'FormattedTime' },
      { exportName: 'IntlProvider' },
    ],
  },
  'react-motion': {
    exportEntries: [{ exportName: 'spring' }],
  },
  'react-router/lib/IndexRoute': {
    exportEntries: [{ exportName: 'default', importAliases: ['IndexRoute'] }],
  },
  'react-router/lib/Redirect': {
    exportEntries: [{ exportName: 'default', importAliases: ['Redirect'] }],
  },
  'react-router/lib/Route': {
    exportEntries: [{ exportName: 'default', importAliases: ['Route'] }],
  },
  urijs: {
    exportEntries: [{ exportName: 'default', importAliases: ['Uri'] }],
  },
  underscore: {
    exportEntries: [
      { exportName: 'clone' },
      { exportName: 'compact' },
      { exportName: 'debounce' },
      { exportName: 'difference' },
      { exportName: 'groupBy' },
      { exportName: 'isEmpty' },
      { exportName: 'isEqual' },
      { exportName: 'isObject' },
      { exportName: 'last' },
      { exportName: 'mapObject' },
      { exportName: 'max' },
      { exportName: 'memoize' },
      { exportName: 'object' },
      { exportName: 'omit' },
      { exportName: 'partial' },
      { exportName: 'pick' },
      { exportName: 'pluck' },
      { exportName: 'range' },
      { exportName: 'shuffle' },
      { exportName: 'sortBy' },
      { exportName: 'tap' },
      { exportName: 'throttle' },
      { exportName: 'times' },
      { exportName: 'values' },
      { exportName: 'without' },
    ],
  },
};
```
Obviously, this is lengthier. I think it could be greatly shortened with some assumptions. For example, if a string is found in the ExportEntries list instead of an object, it could be assumed to represent `{ exportName: <theString> }`.

The main advantage is that the structure allows specifying aliases for any exportName, not just the default. It also allows specification of a real file and whether the file is used for side-effects.

trotzig:
Thanks for sharing your thoughts. I'm humbled and extremely grateful for receiving this type of over-arching, in-depth, feedback. Most of the feedback you get in an open source project is of the type "I need, feature X". I'm still reading through it all and I'll come back to reflect on it after the weekend (I'm afk on a wedding).

rhettlivingston:
You're welcome. Just remember it's more opinionated brainstorming than
anything! Don't be afraid to take it or leave it.

BTW, I'm well on my way to a first cut at an AST-based ES6 import/export
extractor. I state that because I see in its design some intersection with
both JsModule and ImportStatements. It will be interesting to see how it
evolves, assuming it performs.

Enjoy the wedding! Beginnings are an awesome part of life!

lencioni:
I agree with @trotzig! I also won't have time to dig into this today because I am moving. At first glance, I believe you've arrived at many of the same conclusions I did (e.g. options like `useRelativePath` should not have any effect in `JsModule`) which drove me to open this PR in the first place.

rhettlivingston:
I'm OK with the simplification this represents. useRelativePaths is the only one that gives me any pause, but I recognize that moduleNameFormatter delivers a powerful way for the user to do whatever they want with the paths.

My feeling is that if we find we need similar configuration options again for convenience, we'll be at a better place by then to implement them more cleanly.

lencioni:
I agree. I think that we can likely re-implement any of these features in a better way after this simplification, if we want to.

How does this sound:

1. Publish a release of whatever we have in master.
1. Merge this PR.
1. Come up with a list of things we'd like to go into 2.0.
1. Do all of the things on that list.
1. Publish 2.0.0.

In between merging this PR and publishing 2.0.0, we can all use the latest master and shake out any bugs that may be lurking.

rhettlivingston:
Sounds good to me. I'm about to hit "Comment" on a post in #199 that essentially says that I'm heading for a 2.0.0 that makes this look like a drop in the bucket. This likely makes my thoughts become 3.0.0 or we might wait longer than you think for 2.0.0. Either way, still sounds good to me. :smile:

trotzig:
> How to refer to a module in an import statement seems to me to be the job of the ImportStatement class.

I agree. I think it should be fairly easy to move the responsibility. 

> At the highest level of my thoughts, I think there should be a container for JsModule's.

Sounds good too. And just to see if I understand the terminology:

- `bindingIdentifier`: the variable or variables that we are trying to assign an import for.
- `namedExport`: an export by a module that isn't the `default`, e.g. `export { foo, bar }`
- `ExportName`: the name that a `namedExport` is assigned (???)
- `moduleName`: the string we end up using in the import statement, e.g. `components/Button`, `react`, `../../tests/FooTest.js`. 
- `modulePath`: (optional) a path to the file representing the javascript module.

> Problem number one, JsModule doesn't do an adequate job of representing modules we know about but can't find

I think I'm starting to see where you are going. We need to turn `JsModule` from a container class mostly holding information about files into a first-class citizen. We need to keep the container for all `JsModule`s  (`JsModules`?) up-to-date at all times, allowing quick lookup. Something like

```js
const bindingIdentifier = 'React';
const modules = JsModules.find(bindingIdentifier); // Array<JsModule>
```

>  I believe we could get rid of the variableName and include nothing but a namedExports array.

I like that. We'll be closer to the spec, and it's conceptually easier not keeping namedExports and default exports apart. The only class that has to special-case the exports is the ImportStatement class when it's producing its statements. 

> Note that we shouldn't store export aliases. They are a different animal. The export alias ends up being the namedExport and we forget about the internal identifier it aliased.

I'm not completely following the section about aliases/named exports. Can you explain with an example perhaps? 

> Possible problem - I wonder if the filePath (modulePath) should be truly absolute. If not, perhaps lookupPath should be.

This is something I've been thinking about as well. I think it would be safer for us to always use `/absolute/paths`. Right now, we're inconsistent about paths. Most of the times, they're `./local` to the current working directory. Sometimes they're also `local/to` the current working directory. And occasionally they're `/absolute`. I think we can minimize confusion by using absolute paths. This is something we could start doing immediately, so I created an issue: #339. 

Comments on the code block:
`// lookupPaths` - I don't think we need this in JsModule. 


> The main advantage is that the structure allows specifying aliases for any exportName, not just the default. It also allows specification of a real file and whether the file is used for side-effects.

I like how this structure is more expressive, and how it replaces a few (made-up) concepts with one. We can merge this list with what we know about the file system using the same interface. This is what I think you see as the JsModule pool. It's a collection of JsModules coming from different sources. One of those is configuration. One is the file system. One is what we manage to extract from package dependencies.

trotzig:
I'm okay with the plan @lencioni suggested (publishing, merging this PR, moving to 2.0). I've felt a little uneasy about removing some of the configuration options because of how they might make it harder to get started with the project. But I think we are at a better starting point for 2.0 by not having them there. And I think it will get clearer as we move along. 

I'm hoping to get started on making all paths absolute (#339) this week, although there's is a chance I won't have time. We're traveling to the US (San Francisco) on Wednesday.

rhettlivingston:
First, let me warn at what I hinted about on #199 earlier. After sleeping on it, I'm becoming more and more convinced that the EsModule class structure I'm working with for the scanner could grow to replace JsModule and associated things.

>Sounds good too. And just to see if I understand the terminology:
>
> - bindingIdentifier: the variable or variables that we are trying to assign an import for.
> - namedExport: an export by a module that isn't the default, e.g. export { foo, bar }
> - ExportName: the name that a namedExport is assigned (???)
> - moduleName: the string we end up using in the import statement, e.g. components/Button, react, ../../tests/FooTest.js.
> - modulePath: (optional) a path to the file representing the javascript module.

close,,,
- in EcmaScript, default is always a namedExport under the hood (called 'default')
- namedExport is a more casual name that tends to refer to the whole export statement.
- ExportName is a formal name that you can see defined in [Table 42 of the ES7 spec](https://tc39.github.io/ecma262/#table-41)
- In general, absorb tables 39 through 43 from the spec and the terminology's relation to syntax. That will put us all on the same page with the variable names and structures I've used in EsModule and EsModuleModule.

rhettlivingston:
>I think I'm starting to see where you are going. We need to turn JsModule from a container class mostly holding information about files into a first-class citizen. We need to keep the container for all JsModules (JsModules?) up-to-date at all times, allowing quick lookup.

Exactly. I'm currently working on EsModules which is, I think, the beginning of this. It is a container of EsModule objects. EsModule, though, is just a base object that is subclassed by EsModuleModule (an Es6 module found in source). So, other things like predefined EsModules can have their own subclasses to keep the code associated with the different means of loading them separate. And I'd also eventually expect an EsModuleCommonJS subclass because parsing CommonJS is different.

I'm going to go ahead and check in EsModule and EsModuleModule right now so that we have something more concrete to see and discuss. Just realize I've not even run these yet! They lint and flow with no errors, that's it. And it's my first attempt at architecting a serious ES class structure with flow in the loop. I just started learning JS in January! Thankfully, this is far, far from my first language rodeo.

rhettlivingston:
>>Note that we shouldn't store export aliases. They are a different animal. The export alias ends up being the namedExport and we forget about the internal identifier it aliased.
>I'm not completely following the section about aliases/named exports. Can you explain with an example perhaps?
```
// in file 'a'
export { x as y };

// in file 'b'
import { y as z } from 'a';
```
y is the export of file 'a'. 'x' is just a localName aliased by y.
z is the alias of 'a's exported 'y' in file 'b'. This 'z' is what we have been calling an alias in our informal language, The 'y' in file 'a' is a different animal from our normal informal use of the word 'alias'.

rhettlivingston:
> I like how this structure is more expressive, and how it replaces a few (made-up) concepts with one. We can merge this list with what we know about the file system using the same interface. This is what I think you see as the JsModule pool. It's a collection of JsModules coming from different sources. One of those is configuration. One is the file system. One is what we manage to extract from package dependencies.

Exactly! I also see the information in an EsModule as being very concise. The source code and AST are not making it past the constructor - just the digested exports and imports. So we can eventually serialize the harder-to-produce objects to a file system cache and only parse changed files when we start up.

lencioni:
@trotzig you mentioned:

> In our case (brigade), we have an alias that can't be resolved to a file unless we bring in webpack to do the job:

```json
    "branch": "imports?define=>false!branch-sdk",`
```

In that particular case, you should probably make the alias be

```json
    "branch": "branch-sdk",
```

and then configure the imports loader stuff in your webpack config for this module.

PR #331, opened 2016-08-01
Add `importjsd logpath`

lencioni:
When debugging, I need to be able to find the path to the log file. I
think it will be nice to add a command to output this for convenience.


trotzig:
Reasonable. You knew you could just start another instance and look at stdout there, right? This is better though.

PR #332, opened 2016-08-01
Add "Trying local changes" to CONTRIBUTING.md

lencioni:
These tips should help people out who want to contribute to this
project.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,6 +39,20 @@ Using Jest CLI v11.0.2, jasmine2, babel-jest
 285 tests passed (285 total in 7 test suites, run time 29.123s)
 ```
 
+## Trying local changes
+
+You can try your changes out by running `npm link` in the importjs repo
+directory. This will symlink the globally installed version of the package to
+your local version.
+
+To keep the linked version up to date with your changes as you make them, you
+can run `npm run build`. If you'd like to have this automatically happen as you
+make changes, you can run the build process in watch mode: `npm run build --
+--watch`.
+
+When you are done trying your local changes, you can go back to the version you
+had installed by running `npm unlink` in the importjs repo directory.
+
 ## Testing the daemon
 
 This is how you can test the `importjsd` daemon from the command line.


trotzig:
Neat!

rhettlivingston:
I've actually been meaning to ask you guys about this. 

I initially struggled with making changes because I was trying to run the atom importjs addon, make changes to importjs, and try them on a file open in the editor. I linked the .atom/packages/atom-import-js/node_modules/import-js directory over to my import-js clone. That was all very spotty in working for unknown reasons (seemed like it would revert at times, not pick up changes, etc.)

After a while, I got fed up and have been working with an import-js clone that is separated from the atom installation. I've been using node-inspector to debug the command line version of import-js. The process is usually
```bash
cd ../import-js
npm install
cd ../todos
node-debug importjs fix ...somemodule... | sed -e's/\\n/\n/g' | more
```
where, of course, node-debug is actually causing the chrome debugger to pop up in the import-js code so that I can set breakpoints, inspect values, step through, etc.

I also change the package.json "build" entry to add `--source-maps inline` so that node-inspector can let me step through the actual source.

Am I on the right track or have I made some newbie mistakes?

trotzig:
When I've been working on the Atom plugin, I've used `apm link` in the atom-import-js folder along with the regular `npm link` in the import-js folder. To refresh to use the latest import-js changes, @lencioni taught me about bringing up the developer console in Atom, and then hitting cmd+r. It's documented here I believe: https://github.com/Galooshi/atom-import-js/blob/master/CONTRIBUTING.md

As for regular development, I'm either doing something test-driven right now, so all inside the import-js codebase (no editor involved), or I use vim along with the `npm link`ed import-js binaries (restarting vim to get fresh changes). I haven't used the node-debugger much at all to be honest.

rhettlivingston:
> When I've been working on the Atom plugin, I've used apm link in the atom-import-js folder along with the regular npm link in the import-js folder. To refresh to use the latest import-js changes, @lencioni taught me about bringing up the developer console in Atom, and then hitting cmd+r. It's documented here I believe: https://github.com/Galooshi/atom-import-js/blob/master/CONTRIBUTING.md

Thank you. I don't know how I missed that CONTRIBUTING.md. I had discovered the necessity for the apm link instead of npm link, but hadn't figured out the refresh. That probably explains some of the inconsistencies. But, I think you also hinted at my real error when you said "when I've been working on the Atom plugin". I really shouldn't have been trying to debug import-js changes through the atom plugin. It seems to be cleaner to develop it separately.

>As for regular development, I'm either doing something test-driven right now, so all inside the import-js codebase (no editor involved), or I use vim along with the npm linked import-js binaries (restarting vim to get fresh changes). I haven't used the node-debugger much at all to be honest.

Interesting. I do some test-driven development too at which time I'm frequently dribbling console.logs into the code because the debugger can be a pain with jest in the way. But I have gained a lot of insight (maybe of things I didn't want to see at times :smile:) using the source-level, multi-pane debugging that node-inspector brings to the table. I couldn't ask for much more except perhaps being able to make run-time code changes. That tends to be off-the-table or at least complicated in a transpiled environment.

lencioni:
> That tends to be off-the-table or at least complicated in a transpiled environment.

Does it work better if you run `npm run build -- --watch`? That should keep your `npm link`ed importjs updated.

rhettlivingston:
> Does it work better if you run npm run build -- --watch? That should keep your npm linked importjs updated.

Probably not, but I'll have to give it a try sometime just to see what happens. In other languages, I'm used to being able to change a line I'm about to execute and then execute it without saving the change or starting the code over. If I run --watch and "save" the file, I'm betting it will take a reload/restart to see the effect of the change. It will be a learning process to try though :smile:

lencioni:
Ah I see. In Atom, if you have the console open, you can reload Atom by pressing Cmd+r on Mac (I'm guessing Ctrl+r on Windows).

PR #333, opened 2016-08-01
Deprecate .importjs.json

### LINKED ISSUE #326: Deprecate .importjs.json?
We currently support two types of configuration files--JSON and JavaScript. Do we gain much by supporting JSON? Should we deprecate it to simplify people's options?

trotzig:
If you continue to use the JSON configuration format, you'll see a
deprecation message every time you import something:

  Using JSON to configure ImportJS is deprecated and will go away
  in a future version. Use an `.importjs.js` file instead.

Deprecating this now will allow us to remove it in the future, and
hopefully move faster because we have less features to maintain.

[Fixes #326]

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,9 +101,8 @@ the cursor on a variable and hit `<leader>g` (Vim), `(M-x) import-js-goto`
 
 ## Configuration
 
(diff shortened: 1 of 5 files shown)


lencioni:
LGTM, thanks for doing this!

PR #336, opened 2016-08-04
Remove unused code in JsModule#resolvedFilePath

### LINKED ISSUE #335: filePath can't get set to mainFile in JsModule
Noted that this "problem" that was mentioned before in #169 discussions still needs to be fixed. First, to verify that it is needed code and to prevent regression, a test needs to be created that fails on whatever true problem it likely causes.

In jsModule I encountered the following while stepping through code. "startsWith" didn't look at all right.
```js
      if (!this.filePath.startsWith('/package.json')) {
```
I found that it has been in the code since the conversion to JavaScript. Before conversion it was
```js
        return @file_path unless @file_path.end_with?('/package.json')
```
so it's definitely wrong.

trotzig:
We use the resolved file path in the `goto` method, and when we pass in
`pathToImportedModule` to configuration functions. If the file found for
a module is a package.json, we want to use the `main` file when
resolving the "true" file path. When we converted this project from Ruby
to JavaScript, I made a mistake when converting some code handling this.
`startsWith` should have been `endsWith`. We have tests around this, and
they are passing even with the faulty code in place. Which means we are
handling it somewhere else. A quick git-blame led me to 5b9b1031 which
adds code that replaces package.json with the main file:

  if (mainFile) {
    jsModule.filePath = `${importPath}/${mainFile}`;
  }

The fix is to simply remove the faulty code.

[Fixes #335]

### DIFF
--- a/lib/JsModule.js
+++ b/lib/JsModule.js
@@ -154,17 +154,7 @@ export default class JsModule {
     workingDirectory: string = process.cwd()
   ): string {
     if (this.filePath) {
-      // There is a filePath. This happens for JSModules that are not aliases.
-      if (!this.filePath.startsWith('/package.json')) {
-        return this.filePath;
-      }
-
-      if (this.mainFile) {
-        // The filePath points to a package.json file, so we want to look in
-        // that package.json file for a `main` configuration value and open that
-        // file instead.
-        return this.filePath.replace(/package\.json$/, this.mainFile);
-      }
+      return this.filePath;
     }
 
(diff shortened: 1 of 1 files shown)


lencioni:
LGTM

lencioni:
Thanks for the investigation!

rhettlivingston:
Ditto :smile:

trotzig:
Thanks for the reminder @rhettlivingston 👍

PR #338, opened 2016-08-05
Add `importjsd logpath` take 2

lencioni:
My original attempt was reverted by 5c5c3ea7 because it broke the
--parent-pid option. I fixed this by using the commander API a little
better.

This commander package has confusing behavior. Specifically around the
`.command()` function. Apparently if you give it a description argument,
it will treat it like git-style subcommands which means that it looks
for a different file for the action instead of allowing the action to be
defined. And, if you don't specify a description, there isn't a built-in
way to set a command as default via the `isDefault` option. I worked
around this by modifying the arguments that are parsed to set my own
default. Feels a bit hacky, but it works.

### DIFF
--- a/bin/importjsd.js
+++ b/bin/importjsd.js
@@ -1,3 +1,3 @@
 #!/usr/bin/env node
 
-require('../build/daemon.js')();
+require('../build/daemon.js')(process.argv);
--- a/lib/daemon.js
+++ b/lib/daemon.js
@@ -23,80 +23,97 @@ const commandsToFunctionNames = {
   add: 'addImports',
 };
 
-commander.option('--parent-pid <n>', parseInt).parse(process.argv);
+commander.command('start')
+  .description('start the daemon')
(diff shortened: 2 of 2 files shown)


trotzig:
Thanks for taking the time to fix. I agree about the confusing behavior with commander, which is what stopped me from doing a proper fix (instead of the revert). Did you consider filing a bug upstream for that?

lencioni:
From what I can tell, this confusing behavior is intentional. From the documentation:

> When .command() is invoked with a description argument, no .action(callback) should be called to handle sub-commands, otherwise there will be an error. This tells commander that you're going to use separate executables for sub-commands, much like git(1) and other popular tools.
>
> The commander will try to search the executables in the directory of the entry script (like ./examples/pm) with the name program-command, like pm-install, pm-search.

I also found this issue: https://github.com/tj/commander.js/issues/469

trotzig:
Thanks for digging into that.

PR #341, opened 2016-08-20
Update to eslint 3, do some small refactoring

lencioni:
Best reviewed commit-by-commit.

### DIFF
--- a/lib/ImportStatements.js
+++ b/lib/ImportStatements.js
@@ -250,7 +250,7 @@ export default class ImportStatements {
     });
 
     if (groups.length) {
-      groups.filter((group: Array<ImportStatement>): boolean => !!group); // compact
+      groups.filter(Boolean); // compact
     }
     return groups;
   }
--- a/lib/environments/meteorEnvironment.js
+++ b/lib/environments/meteorEnvironment.js
@@ -49,21 +49,27 @@ function meteorPackageDependencies(
   { config }: Object
 ): Array<string> {
   const meteorPackagesPath = path.join(config.workingDirectory, '.meteor/packages');
+
   if (!fs.existsSync(meteorPackagesPath)) {
     return [];
   }
-  // Meteor is an "app" framework. As such, it has both apps and packages. When working with a
-  // module that is part of an app, the list of Meteor packages that a module may import is found in
-  // '.meteor/packages'. This file is actually called a ProjectConstraintsFile in the meteor code.
+
+  // Meteor is an "app" framework. As such, it has both apps and packages. When
+  // working with a module that is part of an app, the list of Meteor packages
+  // that a module may import is found in '.meteor/packages'. This file is
+  // actually called a ProjectConstraintsFile in the meteor code.
   // The internal meteor routine that parses it may be found at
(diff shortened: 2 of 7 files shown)


lib/environments/meteorEnvironment.js
```diff
@@ -308,7 +300,8 @@ function meteorPackageNamedExports({ config }: Object): Object {
   // Try to identify the exports of all packages identified by
   // meteorPackageDependencies
   const namedExports = {};
-  for (const meteorPkg of meteorPackageDependencies({ config })) {
+
+  meteorPackageDependencies({ config }).forEach((meteorPkg: string) => {
```
lencioni: This change was supposed to go in the previous commit, but I accidentally put it here. FYI.

lencioni:
I'm going to merge this so I can keep working. Feel free to review anyway and I'll follow up with any changes.

lib/environments/meteorEnvironment.js
trotzig: cool.

lib/ImportStatements.js
```diff
@@ -250,7 +250,7 @@ export default class ImportStatements {
     });
 
     if (groups.length) {
-      groups.filter((group: Array<ImportStatement>): boolean => !!group); // compact
+      groups.filter(Boolean); // compact
```
trotzig: Thanks for teaching me about this trick!

trotzig:
Looks good to me! Thanks @lencioni!

lib/environments/meteorEnvironment.js
```diff
@@ -276,24 +276,16 @@ function extractExportsFromMeteorPackage(
...
-      // collection.
-      declaredExports.push(declaredExport.name);
     });
 
     // TODO: If the "resources" section of the buildIsopack specifies a
     // mainModule, we need to attempt to scan it to find exports.
   });
 
-  return declaredExports;
+  return Array.from(declaredExports);
```
rhettlivingston: In trying to use Sets more myself, I ran into the problem that Flow has a bug causing it to errantly complain about a type mismatch with the normal [...declaredExports] syntax. I'm guessing you ran into it too. I've found that an explicit cast gets around the problem, i.e. `[...(declaredExports: any)]`. A true fix is in the works in Flow.

lencioni: Oh, good to know. I actually hadn't run into that.

trotzig: This bug? https://github.com/facebook/flow/issues/1059#issuecomment-241267694

I'm a happy subscriber of that issue. :)

rhettlivingston: Yep, though I've been following a PR that fixes it - facebook/flow#1668.

That syntax seems to be what the community is converging on as most common for the conversion. It's also used in most examples of how to work around the lack of a built-in union operator,,, like [this one](http://www.2ality.com/2015/01/es6-set-operations.html).

As Pascal was my first "real" language (circa 1982 after Cobol, Basic, and some assembly), I learned to appreciate sets early on, especially in language processing applications.  You'll probably see me using more of them.

PR #342, opened 2016-08-20
Remove deprecated local config and snake_cased config handling

lencioni:
These have been deprecated and we expect our next release to be a major version bump, so let's jettison these lines of code and move forward to 2.0 with a leaner codebase.


trotzig:
👍

PR #346, opened 2016-08-24
Use `const` as declaration keyword in node environment

### LINKED ISSUE #235: Node env configuration should use `const` as declaration keyword by default

lencioni:
This is typical for the node environment, and likely will be until node
gets `import`s. People who are transpiling their code in node can still
override this by setting `declarationKeyword` in their configuration.

Fixes #235

### DIFF
--- a/lib/__tests__/Importer-test.js
+++ b/lib/__tests__/Importer-test.js
@@ -659,20 +659,16 @@ Meteor
         });
       });
 
-      describe('in a node environment', () => {
-        beforeEach(() => {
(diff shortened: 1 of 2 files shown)


trotzig:
Looks good. You should update the configuration for this project so that it continues to use `import`.

trotzig:
Is this still valid @lencioni ?

lencioni:
Uh, yeah I think so. Looks like `import` is close to landing in node though it isn't there quite yet.

PR #347, opened 2016-08-24
Default importjsd to --help instead of start

### LINKED ISSUE #345: `importjsd --parent-pid n` errors
I think I actually broke this when I thought I had fixed it. Workaround is to run `importjsd start --parent-pid n` which I've done e.g. for the Vim plugin here: https://github.com/Galooshi/vim-import-js/commit/3103f2e12674d81ce15daeb7db2ab62c768a4390

At this point, I think the best solution is to update the plugins to use `importjsd start` and stop trying to default `importjsd` to `importjsd start`.

lencioni:
This currently does not work correctly when used like `importjsd
--parent-pid n`, which has caused some breakages in our editor plugins.
Since this does not seem to be a thing that commander.js likes to do
very easily, I think the best path forward at this point is to simply
default to `--help` instead.

I've already updated the editor plugins to use `importjsd start
--parent-pid n`.

Fixes #345

### DIFF
--- a/lib/daemon.js
+++ b/lib/daemon.js
@@ -109,11 +109,10 @@ commander.command('logpath')
   });
 
 export default function daemon(argv: Array<string>) {
-  const args = argv;
-  if (args.length <= 2) {
-    // no arguments were given, so default to start
-    args.push('start');
-  }
+  commander.parse(argv);
 
-  commander.parse(args);
+  if (argv.length <= 2) {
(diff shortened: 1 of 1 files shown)


trotzig:
Thanks Joe! Will you take care of releasing the sublime plugin?

trotzig:
I see you already did. Thanks!

PR #358, opened 2017-01-03
Es7 features

trotzig:


lencioni:
Decorators and class properties are not es7. They are still proposals that
have not yet hit stage 3.

Issue #337, opened 2016-08-05
Watchman integration giving wrong results when parent directory is in watch-list


lencioni:
I haven't had time to fully track this down yet, but I've been noticing some odd behavior when using import-js with watchman integration.

Specifically, I keep my repos in my home directory. At a fresh start, things are working just fine.

At some later point, I notice that the results that import-js is finding aren't correct. Instead of something like "../path/to/thing" the module name is more like "../../project/path/to/thing". This is additionally weird because that path isn't even correctly relative to the current file--in this case it should be "../../../project/path/to/thing".

And sometimes, after this point, when I run fix imports, vim ends up locking up and taking 100% CPU. The only way to recover is to `kill -9` the process. Import word does not cause vim to have this CPU usage and freeze, but it does have the strange paths. This might be a separate issue.

I've noticed that my home directory appears in the watch-list when this happens. I'm not entirely sure why, or why it might have an effect. Also, the 100% CPU usage may be unrelated (I also happen to be running the remove-strip-from-path branch locally, which may be part of this). 

```
~/project ❯❯❯ watchman watch-list
{
    "version": "4.6.0",
    "roots": [
        "/Users/joe_lencioni",
        "/Users/joe_lencioni/project",
    ]
}
```

If I remove my home directory from the watch list, the paths start working correctly again. However, I still end up with 100% CPU sometimes (which makes me further think that is a separate issue).

lencioni:
I wonder if any of this has anything to do with me breaking `importjsd --parent-pid #`

trotzig:
I doubt it has anything to do with that. But there might have been a bug at some point causing the root folder to be hooked up. 

Here are the folders I watch currently:

```
{
    "version": "4.5.0",
    "roots": [
        "/Users/henrictrotzig/react-simple-pie-chart",
        "/Users/henrictrotzig/vim-import-js",
        "/Users/henrictrotzig/brigade",
        "/Users/henrictrotzig/causes",
        "/Users/henrictrotzig/import-js"
    ]
}
```

Have you tried restarting the watchman service?

trotzig:
We resolved this offline. The issue was (if I recall correctly) that watchman had reached its match file limit, because of how the home folder was watched. 

Since this was reported, import-js has code in place to prevent enabling watchman for certain folders, which should prevent the home folder from getting watched.

lencioni:
Specifically, I think this was fixed by 0f9ed418b3d27e2eefac3936d45733c04a37f2fa.

[...]

PR #365, opened 2017-01-20
Find named exports

trotzig:
This ended up taking a little longer than I was hoping for. I've been slowly iterating from a version of import-js that traverses the file system on every import into a version that keeps an up-to-date cache/index of modules. I'm finally in a stable (ish) place. 😃 

I don't expect anyone to look at all these changes. I've published a beta version (2.1.0-beta.5) that you can try out. 

The most important files added/changed are:
- ModuleFinder
- ExportsStorage
- findExports

Ping @rhettlivingston, @lencioni - if either of you have time to review, I'd be thrilled. And @rhettlivingston, since this is a complete overhaul of named exports, it's likely that I might have broken something in the meteor environment. It could also be that you are able to make use of the new system to make finding named meteor packages simpler.


lib/ModuleFinder.js
```diff
@@ -0,0 +1,216 @@
...
+        excludes,
+        ignorePackagePrefixes,
+      });
+      instances[workingDirectory] = instance;
+    }
+    return instance;
+  }
+
+  constructor(workingDirectory, { excludes, ignorePackagePrefixes }) {
+    this.excludes = excludes;
```
lencioni: This might be fine, but I think in general the more we can avoid state, the better--and I see a lot of state in this class.

lib/Watcher.js
```diff
@@ -0,0 +1,184 @@
...
+      }, (error: Error) => {
+        if (error) {
+          client.end();
+          reject(error);
+        } else {
+          this.startSubscription({ client }).then(resolve).catch(reject);
+        }
+      });
+    });
+  }
```
lencioni: note to self: this is where I stopped reviewing for now.

README.md
```diff
@@ -601,4 +601,10 @@ as a daemon).
 See the [CONTRIBUTING.md](CONTRIBUTING.md) document for tips on how to run, test
 and develop ImportJS locally.
 
+## Thank you:
+- @janpaul123 for writing the Sublime plugin.
```
trotzig: Ah, I made a guess that this would work. I'll add the links back.

bin/importjsd.js
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env node
 
-require('../build/daemon.js')(process.argv);
+// The `importjsd` command is deprecated and will be removed in a future
+// version. `importjs` should be used instead.
+require('../build/importjs.js')(process.argv);
```
trotzig: agreed, though this is probably fine since we're deprecating.

lib/ExportsStorage.js
```diff
@@ -0,0 +1,241 @@
...
+        if (err) {
+          reject(err);
+          return;
+        }
+        resolve();
+      });
+    });
+  }
+
+  needsUpdate(files) {
```
trotzig: Yeah, it's hard to step away from the object-oriented past self. I'll think about this some more, at least this file doesn't have that much state (`ModuleFinder` is worse).

lib/ModuleFinder.js
trotzig: Yeah. If you can think of ways to make this better, drop a line or so here. One of the reasons state is needed is because we have long-running instances, and the call sites usually don't have enough state themselves to pass in the right things.

lib/Watcher.js
trotzig: This is where you had enough already. ;)

lib/ModuleFinder.js
```diff
@@ -0,0 +1,223 @@
...
+  }
+  if (dirName === '.') {
+    return [fileName];
+  }
+  const result = [
+    `${dirName}-${fileName}`,
+    fileName,
+  ];
+
+  if (/e?s$/.test(dirName)) {
```
lencioni: A comment might not be a bad idea around here to explain why this is in here.

lib/ModuleFinder.js
```diff
@@ -0,0 +1,223 @@
...
+        this.processQueue(done);
+      });
+    }).catch((err) => {
+      logLevel.error(err.message);
+      this.processQueue(done);
+    });
+  }
+
+  handleFilesAdded(unexpandedFiles) {
+    return new Promise((resolve) => {
```
lencioni: With all of these promises, it might be worth spending some time going through and making sure that errors aren't being swallowed.

lib/Watcher.js
```diff
@@ -0,0 +1,184 @@
...
+        }).then(resolve).catch(reject);
+      });
+    });
+  }
+
+  initialize(): Promise<void> {
+    return new Promise((resolve: Function, reject: Function) => {
+      this.initializeWatchman().then(resolve).catch(() => {
+        loglevel.info(
+          "Couldn't initialize watchman watcher. Falling back to polling.");
```
lencioni: This might happen elsewhere, but it might be nice to provide more information about why unable to init watchman.

lib/Watcher.js
```diff
@@ -0,0 +1,184 @@
...
+          reject(error);
+        } else {
+          this.startSubscription({ client }).then(resolve).catch(reject);
+        }
+      });
+    });
+  }
+
+  initializePolling(): Promise<void> {
+    setInterval(() => {
```
lencioni: Should this have some sort of a shutdown mechanism that disables this interval?

lib/__tests__/ModuleFinder-test.js
```diff
@@ -0,0 +1,94 @@
...
+  requireResolve.__addResolvedPath(
+    path.join(process.cwd(), 'node_modules/react-packer'),
+    path.join(process.cwd(), 'node_modules/react-packer/index.js'),
+  );
+
+  moduleFinder = ModuleFinder.getForWorkingDirectory(process.cwd(), {
+    excludes: [],
+    ignorePackagePrefixes: ['react-'],
+  });
+  return moduleFinder.initializeStorage(':memory:');
```
lencioni: I'm not the biggest fan of having a bunch of things happen in beforeEach, because it makes each individual test more difficult to reason about and it can make it more difficult to add new tests to the file. I think this test is maybe small enough and the setup is maybe large enough so that it is okay, but it might be something worth considering inlining into each test for maximum clarity.

lib/daemon.js
```diff
@@ -23,71 +23,68 @@ const commandsToFunctionNames = {
...
+    const importer = new Importer(
+      payload.fileContent.split('\n'),
+      payload.pathToFile,
+      workingDirectory
+    );
 
+    const functionName = commandsToFunctionNames[payload.command];
+    if (!functionName) {
+      const errorString =
+        `Unknown command: ${payload.command}. ` +
```
lencioni: I think we shouldn't worry about breaking up long strings via concatenation

lib/findAllFiles.js
```diff
@@ -0,0 +1,23 @@
+import path from 'path';
+
+import glob from 'glob';
+
+import lastUpdate from './lastUpdate';
+
+export default function findAllFiles(workingDirectory, excludes) {
+  return new Promise((resolve, reject) => {
+    glob('./**/*.js*', {
```
lencioni: I wonder if there is a more specific glob pattern that we could use. Like, does this work? 
```
.*/**/*.@(js|jsx|json)
```

https://github.com/isaacs/node-glob#glob-primer

lib/findExports.js
```diff
@@ -0,0 +1,203 @@
+import fs from 'fs';
+import path from 'path';
+
+import parse from './parse';
+import requireResolve from './requireResolve';
+
+function findESNamedExports(node) {
+  if (node.type !== 'ExportNamedDeclaration') {
```
lencioni: Some of these AST bits could still probably benefit from comments with examples of what they mean.

lib/findExports.js
```diff
@@ -0,0 +1,203 @@
...
+    if (!left.object || !left.property) {
+      // foo = 'bar';
+      return false;
+    }
+    return left.object.name === 'module' &&
+      left.property.name === 'exports';
+  });
+}
+
+const DEFAULT_EXPORT_PATTERN = /\smodule\.exports\s*=\s*(\w+)/;
```
lencioni: If you start this pattern with `\s`, will it match if `module.exports` is on the first line without any whitespace before it? You might need to tweak this pattern.

lib/findJsModulesFor.js
```diff
@@ -22,71 +19,69 @@ function findImportsFromEnvironment(
...
-        const isExcluded = excludes
-          .some((globPattern: string): boolean => minimatch(f, globPattern));
-        if (isExcluded) {
-          return;
-        }
+    let isWantedPackageDependency = Boolean;
+    if (!config.get('importDevDependencies')) {
+      const packageDependencies = config.get('packageDependencies');
+      isWantedPackageDependency = (packageName: string): boolean =>
+        packageDependencies.indexOf(packageName) !== -1;
```
lencioni: We should make `config.get('packageDependencies')` return a `Set` to avoid this array scan.

lib/findProjectRoot.js
```diff
@@ -0,0 +1,27 @@
+import fs from 'fs';
+import path from 'path';
+
+function findRecursive(directory) {
+  if (directory === '/') {
+    throw new Error('No project root found');
+  }
+  const pathToPackageJson = path.join(directory, 'package.json');
+  const pathToNodeModulesFolder = path.join(directory, 'node_modules');
+  const isPackageDependency = /\/node_modules$/.test(path.dirname(directory));
```
lencioni: Move the pattern to a constant outside of this function.

lib/importjs.js
```diff
@@ -98,20 +116,49 @@ program.command('add <imports> <pathToFile>')
...
+program.command('cachepath')
+  .description('show path to cache file')
+  .action(() => {
+    stdoutWrite(new Configuration('importjs').get('cacheLocation'));
+  });
+
+program.command('logpath')
+  .description('show path to log file')
+  .action(() => {
+    process.stdout.write(`${pathToLogFile}\n`);
```
lencioni: Should we use `stdoutWrite` here instead?

lib/visitIdentifierNodes.js
```diff
@@ -1,12 +1,34 @@
...
+      parent.parent.type === 'JSXOpeningElement')) {
+      return {
+        name: node.name,
+        isJSX: true,
+        definedInScope,
+      };
+    }
+  }
+
+  if (parent.type === 'GenericTypeAnnotation') {
```
lencioni: Some AST example comments here would be nice too

lib/ModuleFinder.js
trotzig: Good idea. Done in 9a6dcc9

lib/ModuleFinder.js
trotzig: I've been running this code quite intensely for a month+ now. I hope to have found all the places where we're not handling rejections.

lib/Watcher.js
trotzig: True. Done in 8c730b3fd

lib/Watcher.js
trotzig: It should run as long as the process is active. 

To be honest, I'm considering cutting out the polling completely. It's hard to test, and won't provide a good user-experience. What do you think? Hard dependency on watchman? It runs on most os:es.

lib/__tests__/ModuleFinder-test.js
trotzig: I agree with you in general, but I also think it's fine in this one case. The `Importer` test is an example of where this has gone way out of proportion (sigh).

lib/daemon.js
trotzig: Ah, you mean just 

```js

const errorString = `
  Unknown command: ${payload.command}. 
  Valid ones are ${Object.keys(commandsToFunctionNames).join(', ')}
`.trim();
```

?

lib/findExports.js
trotzig: I should have done them while I was adding them... Some of them I can't fully recall why they're there. Any addition touching the ast has been very TDD, so hopefully there are good info if you start git-blaming.

lib/findJsModulesFor.js
trotzig: Good call. Done in f41b247.

lib/findProjectRoot.js
trotzig: Done in a9803f6

lib/importjs.js
trotzig: Done in 2147514

lib/visitIdentifierNodes.js
trotzig: Yeah... I'll add to my todo to make a sweep here.

lib/daemon.js
lencioni: That would be okay, but then you get a bunch of extra spaces in the middle. I was thinking more like this:

```js
const errorString = `Unknown command: ${payload.command}. Valid ones are ${Object.keys(commandsToFunctionNames).join(', ')}`;
```

But it doesn't really matter a whole lot

lib/Watcher.js
lencioni: That makes sense to me. It might be worth considering making this pluggable, so you can use watchman or swap it out for some other thing if you want (e.g. polling, fsevents, or some other file watcher).

lib/findExports.js
lencioni: I see. What if it is wrapped in parens like this?

```js
(module.exports = Foo)
```

trotzig: We could fix that, but I don't know if it's worth it. There are numerous other cases where the regex won't work (line-breaks, not exporting an identifier, curly braces, etc). It should be seen as a nice-to-have, but not strictly useful.

trotzig:
The code climate check fails because I'm more than 50 commits ahead of master. https://docs.codeclimate.com/docs/analysis-error-codes#G12

Going to ignore that for now and hope that things return to normal once on master.

lencioni:
I'm using beta 7 locally and noticed that it doesn't find `expect` from chai. https://github.com/chaijs/chai#native-modules-usage-local-import-only

I haven't dug into why, but I figured I'd point it out for you.

trotzig:
That's good feedback! I'll look into why the exports aren't found. 

I made a decision not to reach into package dependencies folders. I think it's considered bad practice to expose internal structure from a package ([React is moving away from it](https://github.com/facebook/react/issues/8854#issue-202694221)). I think the best thing we can do is to file issues/PRs to make these projects use flat bundles.

trotzig:
I fixed the chai issue. The other one I wasn't able to reproduce. `npm install --save enzyme` made it possible to import `shallow`. Maybe it's an old version? Can you try `rm $(importjs cachepath)` and then try again? One annoying part of that is that I made the watcher initialize lazily, so it won't start until the first time you try importing something. At which time the cache is empty, so you won't get results the first time.

lib/Watcher.js
```diff
@@ -123,7 +123,7 @@ export default class Watcher {
         this.subscribe({
           client,
           fbWatch: resp.watch,
-          relativePath: resp.relativePath,
+          relativePath: resp.relative_path,
```
lencioni: I believe this fixed #337

PR #366, opened 2017-02-03
Make it clearer that 'memoize' is a namedExports

asfktz:
Even though that feature is deprecated, it might be more clear to see that `memoize` was defend as `namedExports`
before we use it on the next example (`import { memoize } from 'underscore';`)


asfktz:
Sure 😊

PR #367, opened 2017-02-09
Bugfixes

trotzig:
I found a few issues with the 2.1.0 release, mostly related to the polling mechanism of Watcher.

### DIFF
--- a/lib/Watcher.js
+++ b/lib/Watcher.js
@@ -117,7 +117,7 @@ export default class Watcher {
 
   initialize(): Promise<void> {
     return new Promise((resolve: Function, reject: Function) => {
-      this.initializeWatchman().then(resolve).catch(() => {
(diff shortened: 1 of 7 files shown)


lib/Watcher.js
```diff
@@ -126,6 +126,8 @@ export default class Watcher {
 
           ImportJS will fall back to polling instead. This is slower, and more
           resource-heavy. Plus, changes will not be immediately available.
+          \n
+          Reason: ${error.message}\n${error.stack}
```
lencioni: I think `error.stack` might include `error.message`, so this might be redundant, but NBD. (Although I don't think `error.stack` has been standardized yet, so maybe worth having both anyway I guess.)

trotzig: Cool, I didn't know.

PR #370, opened 2017-02-13
Fix support for namespace imports

### LINKED ISSUE #369: `Fix all imports` removes star imports (* as)
Hi,
how are you? 😊

First, I want to thank you so much for the 2.0 release,
Auto import named exports is amazing!

Lately, I encountered a regression where calling `importjs: fix all imports` removes all import expression with `* as` syntax

I'm not 100% sure, but I believe that it starts with `v2.0.0`, where earlier versions (`v1.1.2`) works fine.

Steps to reproduce:
1. Create 2 files:

```js
// api.js

export function get () {
  // ....
}

export function post () {
  // ....
}
```

```js
// index.js

import * as api from './api'

api.get()
```

2. Call `importjs: fix all imports` on index.js

Results:
```js
// index.js

api.get()
```

Tested on:
- Sublime Text 3
- node `v6.6.0`
- importjs `2.1.0`
- macOS sierra

trotzig:
When running fix imports on something that looked like this

  import * as foo from 'foo';

...we would incorrectly remove the import. This was because we weren't
properly identifying the "as foo" part as an assignment.

Since I usually don't use imports of this type, I worry that we'll
regress in the future. Even though the fix could have been isolated to
findCurrentImports, I decided to flesh out a few other tests just to
make sure we have better coverage for the fix.

Fixes #369. Thanks to @asfktz for the report!

### DIFF
--- a/lib/__tests__/Importer-test.js
+++ b/lib/__tests__/Importer-test.js
@@ -670,9 +670,9 @@ foo
 
           it('adds the import and sorts the entire list with groups', () => {
             expect(subject()).toEqual(`
-import * as starBar from './star/bar';
 import bar from './foo/bar';
 import foo from './bar/foo';
+import * as starBar from './star/bar';
 import zoo from './foo/zoo'
 
 const sko = customImportFunction('./sko');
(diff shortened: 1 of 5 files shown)


lib/__tests__/Importer-test.js
```diff
@@ -1965,6 +1965,25 @@ foo();
       });
     });
 
+    describe('with unconventional imports', () => {
```
lencioni: How about "namespace" instead of "unconventional" here?

trotzig: Done, thanks!

PR #371, opened 2017-02-15
Pick first match for goto

trotzig:
When using `goto` and the word couldn't be matched to a single file, we
would send back a list of `unresolvedImports` to the editor plugin.
However, none of the editor plugins (sublime, vim, emacs, atom) knows
how to resolve ambiguities for goto. In most cases, the goto path can be
resolved using current imports. In the rare event that it can't be
resolved, we now just pick the first one in the list.

This is a quick-fix to a problem that could be address more
holistically. But the more involved fix would include updating all the
plugins, and that's something I'd like to avoid right now.

### DIFF
--- a/lib/Importer.js
+++ b/lib/Importer.js
@@ -98,8 +98,16 @@ export default class Importer {
         variableName,
         this.pathToCurrentFile
       ).then((jsModules: Array<JsModule>) => {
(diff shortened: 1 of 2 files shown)


trotzig:
There is now! https://github.com/Galooshi/import-js/issues/372

PR #377, opened 2017-02-22
Exclude later

trotzig:
I ended up pushing two commit here. My first attempt was simply to apply yet another exclusion filter. This made things a little better. But after thinking about implications, I decided to push another more aggressive change that consolidated all the exclusion filtering into one place -- in `findJsModulesFor` at import-time.

### DIFF
--- a/lib/findJsModulesFor.js
+++ b/lib/findJsModulesFor.js
@@ -1,5 +1,6 @@
 // @flow
 
+import minimatch from 'minimatch';
 import sortBy from 'lodash.sortby';
 import uniqBy from 'lodash.uniqby';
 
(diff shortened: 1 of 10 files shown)


lib/findJsModulesFor.js
```diff
@@ -48,6 +49,13 @@ function findJsModulesFromModuleFinder(
             hasNamedExports: !isDefault,
           });
         }
+
+        // Filter out modules that are in the `excludes` config.
+        if (config.get('excludes').some((glob: string): boolean =>
+          minimatch(path, glob))) {
```
lencioni: nit: I'm finding the indentation here to be a little confusing. It might be better to break this up in a different way--e.g. storing something in a variable).

trotzig:
Ouch, that's a lot of excluded files. May I ask what those are? 

I'll think about ways to react to config changes. An immediate idea would be to hash together something based on certain config options (where `excludes` is one of them), and then use that hash to invalidate the cache.

lencioni:
These are the excludes with globs I am using:

```js
'./coverage/**',
'./public/**',
'./tmp/**',
```

most of the files (~23,000) are in tmp.

trotzig:
Oh yeah, that's significant. I think this PR (as-is) isn't the right solution then.

PR #382, opened 2017-02-26
Update dependencies, reformat code with prettier

lencioni:


dcporter:
(I'm looking in from the outside so please weigh accordingly, but my experience has been that the trailing commas rule has caused much more strife and developer frustration than it's saved in commit diff noise.)

lencioni:
@dcporter are you referring to requiring trailing commas on multiline function arguments or trailing commas in general?

dcporter:
Right, this PR (because airbnb I think?) is doing it with the function calls; my experience is with that and with multiline object definitions.

dcporter:
On the other hand there is great value in attaching to an existing thing and not nitpicking it, and airbnb's linter seems to be getting pretty popular, so YMMV =)

trotzig:
Any ideas why code coverage went down by -7.4%?

lencioni:
Greenkeeper makes sense to me. Not sure why coverage went down.

PR #388, opened 2017-03-06
Delay visiting `body` arrays in findUndefinedIdentifiers

### LINKED ISSUE #383: findUndefinedIdentifiers returns anything that's referenced before it's declared.
If a file references a variable higher in the file than that variable's declaration — valid JS if the reference is within a function — `findUndefinedIdentifiers` flags it as not locally defined. This can result in bad imports, especially invalid same-file imports like `import { abc } from [this same file which exports abc]`. (The same-file issue is where I ran into it, but it'll add the import if that symbol is exported from anywhere.)

In this minimal example, `abc` will be returned from `findUndefinedIdentifiers` despite being defined (and, since `abc` is an export, it will be added as an import to this same file on `importjs fix`):

```js
function modifyAbc() {
    abc.def = 4;
}

export const abc = {};
```

The root cause is that `visitIdentifierNodes` ([here](https://github.com/Galooshi/import-js/blob/master/lib/visitIdentifierNodes.js#L57)) walks the AST tree once, looking for declarations and references *and comparing them* in the same pass. Solutions include walking the tree twice, or walking it once but compiling two lists (references and declarations) and comparing them at the end.

(Followup: the fix may be more complicated than that. I'm having trouble tracing data through those calls, and scopes are hard.)

trotzig:
When a variable is defined after it being used inside a function scope,
`findUndefinedIdentifiers` would incorrectly flag the variable as
undefined. Here's an example:

  function foo() {
    console.log(bar);
  }

  const bar = 'fake news';

As you can see, `bar` is defined after the function making use of it.
This is perfectly valid javascript, and executing `foo()` would log
`"fake news"` to the console. To take this scenario into account, we had
to delay visiting nodes inside a function body until the parent scope
was "done" (meaning we've found all defined identifiers in it).
Essentially, this meant switching from a depth-first tree traversal to a
breadth-first. I couldn't think of a better way to do that than by
switching from the recursive version that we had before into an iterativ
solution.

This fix ended up being strongly TDD'ed. I'm not completely happy with
where the code ended up, but for now I can't come up with a more elegant
solution.

Fixes #383

### DIFF
--- a/lib/__tests__/findUndefinedIdentifiers-test.js
+++ b/lib/__tests__/findUndefinedIdentifiers-test.js
@@ -164,6 +164,22 @@ it('knows about objects', () => {
   ).toEqual(new Set(['foo', 'uuid']));
 });
 
+it('knows about identifiers declared after usage', () => {
+  expect(
+    findUndefinedIdentifiers(
+      parse(
+        `
+    function modifyAbc() {
+      abc.def = 4;
+    }
+
+    export const abc = {};
+  `,
+      ),
+    ),
+  ).toEqual(new Set([]));
+});
+
 it('handles es6 imports + commonjs + react', () => {
   expect(
     findUndefinedIdentifiers(
--- a/lib/visitIdentifierNodes.js
+++ b/lib/visitIdentifierNodes.js
@@ -56,37 +56,59 @@ function normalizeNode(node, { key, definedInScope, parent }) {
 }
 
 export default function visitIdentifierNodes(
-  node,
+  rootAstNode,
   visitor,
   context = { definedInScope: new Set([]), key: 'root' },
 ) {
-  if (Array.isArray(node)) {
-    node.forEach((child) => {
-      visitIdentifierNodes(child, visitor, context);
-    });
-    return;
-  }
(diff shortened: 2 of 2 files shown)


lib/__tests__/findUndefinedIdentifiers-test.js
```diff
@@ -164,6 +164,22 @@ it('knows about objects', () => {
...
+  expect(
+    findUndefinedIdentifiers(
+      parse(
+        `
+    function modifyAbc() {
+      abc.def = 4;
+    }
+
+    export const abc = {};
+  `,
```
trotzig: @lencioni I noticed that `prettier` made the indentation of these strings odd. I followed local formatting, but I'd like to hear your thoughts, perhaps it's a bug in prettier?

lencioni: I think it is intentional because they don't want to modify the contents of the strings.

trotzig: Yeah, I had to think about this a few times before figuring it out. Makes sense.

PR #394, opened 2017-04-11
Find exports defined through Object.defineProperty

trotzig:
I came across a package (graphql-relay) where import-js wouldn't find
any named exports, despite it having a few listed. Exports were defined
like this:

    Object.defineProperty(exports, 'foo', {
      enumerable: true,
      get: function get() {
        return 'foo';
      }
    });

While working on the fix, I made sure that we cover shadowed variables
for `exports` as well.


trotzig:
Fixes #394

PR #395, opened 2017-04-12
Improve "No project root found" error message, simplify project root check to only look for package.json

lencioni:
This error message is too vague and can be confusing for folks. We can
make it better by providing some more context.

More info: https://github.com/Galooshi/atom-import-js/issues/13

---

Stop checking for node_modules when finding project root

In https://github.com/Galooshi/import-js/pull/395 we discussed the logic
for determining that something is a project root. It seems that there
are situations where you might have a package.json file but no
node_modules directory (e.g. haven't run `npm install` yet, or there are
no dependencies). I think we can improve this by simplifying the check
to only look for package.json.

### DIFF
--- a/lib/__tests__/findProjectRoot-test.js
+++ b/lib/__tests__/findProjectRoot-test.js
@@ -25,7 +25,7 @@ it('treats package folders as roots', () => {
 
 it('throws if it can not find a folder', () => {
   expect(() => findProjectRoot('/foo/bar/baz.js')).toThrowError(
-    'No project root found',
+    'No project root found, looking for a directory with either a package.json file or a node_modules directory.',
   );
 });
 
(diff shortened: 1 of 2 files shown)


lencioni:
What do you mean by "local packages hidden deep in the tree structure"? Can you give me an example of this?

trotzig:
Something like

```
project
  - package.json
  - node_modules
    - ...
  - src
    - foo
      - package.json
      - foo-main.js
    - bar
```

The `foo` module here, specifying its main file as foo-main.js.

Never seen it in the wild, but it's valid, right?

lencioni:
Ah I see. A sub-package that has no dependencies but has other package behavior. Yeah I could see that happening in monorepos. This could even happen if you just haven't run npm install in that directory yet. In any case, I think the existence of a package.json indicates the project root.

trotzig:
Will you make that change? (Check for package.json only). 

The error message in this PR is still better than what is there currently, so I'm fine merging.

lencioni:
@trotzig PR updated

PR #398, opened 2017-05-15
Use path.isAbsolute for Windows compatibility

### LINKED ISSUE #397: Use path.isAbsolute for Windows compatibility
### Issue

The [`makeAbsolute`](https://github.com/Galooshi/import-js/blob/master/lib/findProjectRoot.js#L18) function already receives an [absolute path](https://nodejs.org/api/path.html#path_path_isabsolute_path) ("C:\path") on Windows but only checks if it begins with '/' and so wrongly tries to convert it with the `path.join` where `process.cwd()` is actually "C:\Users\<username>\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\decompress-zip\node_modules\graceful-fs\polyfills.js", turning `pathToFile` into a mish-mash of paths.

For some reason this then causes `findRecursive` to enter an infinite loop.

### Fix 

Replace `pathToFile.startsWith('/')` with `path.isAbsolute`

____________________

Steps to reproduce:

1. Activate any of the 3 commands
 
**Atom**: 1.16.0 ia32
**Electron**: 1.3.13
**OS**: Microsoft Windows 7 Professional (64-bit)
**Thrown From**: [atom-import-js](https://github.com/galooshi/atom-import-js) package 0.6.1

### Stack Trace

Uncaught RangeError: Maximum call stack size exceeded

```cmd
At path.js:64

RangeError: Maximum call stack size exceeded
    at String.slice (native)
    at normalizeStringWin32 (path.js:64:22)
    at Object.normalize (path.js:403:14)
    at Object.join (path.js:527:18)
    at findRecursive (/packages/atom-import-js/node_modules/import-js/build/findProjectRoot.js:28:10)
    (then the above line repeated as many times as possible)
    at findProjectRoot (/packages/atom-import-js/node_modules/import-js/build/findProjectRoot.js:39:10)
    at withModuleFinder (/packages/atom-import-js/lib/withModuleFinder.js:6:26)
    at importWord (/packages/atom-import-js/lib/import-js.js:112:3)
    at /packages/atom-import-js/lib/import-js.js:150:33)
    at CommandRegistry.module.exports.CommandRegistry.handleCommandEvent (/app.asar/src/command-registry.js:259:29)
    at CommandRegistry.handleCommandEvent (/app.asar/src/command-registry.js:3:59)
    at CommandRegistry.module.exports.CommandRegistry.dispatch (/app.asar/src/command-registry.js:160:19)
    at AtomEnvironment.module.exports.AtomEnvironment.dispatchContextMenuCommand (/app.asar/src/atom-environment.js:1185:28)
    at EventEmitter.outerCallback (/app.asar/src/application-delegate.js:325:25)
    at emitThree (events.js:116:13)
    at EventEmitter.emit (events.js:194:7)
```

reohjs:
Replaced `pathToFile.startsWith('/')` with `path.isAbsolute`

Closes #397

### DIFF
--- a/lib/normalizePath.js
+++ b/lib/normalizePath.js
@@ -1,4 +1,5 @@
 // @flow
+import path from 'path';
 
 export default function normalizePath(
   rawPath: ?string,
--- a/lib/normalizePath.js
+++ b/lib/normalizePath.js
@@ -1,5 +1,4 @@
 // @flow
-import path from 'path';
 
 export default function normalizePath(
   rawPath: ?string,
--- a/lib/normalizePath.js
+++ b/lib/normalizePath.js
@@ -17,7 +17,7 @@ export default function normalizePath(
   }
 
   if (
-    rawPath === '.' || rawPath.startsWith('./') || rawPath.startsWith('../')
+    rawPath === '.' || rawPath.startsWith('./') || rawPath.startsWith('../') || path.isAbsolute(rawPath)
   ) {
     return rawPath;
   }
--- a/lib/normalizePath.js
+++ b/lib/normalizePath.js
@@ -18,7 +18,7 @@ export default function normalizePath(
   }
 
   if (
-    rawPath === '.' || rawPath.startsWith('./') || rawPath.startsWith('../') || path.isAbsolute(rawPath)
+    rawPath === '.' || rawPath.startsWith('./') || rawPath.startsWith('../')
   ) {
     return rawPath;
   }
--- a/lib/Importer.js
+++ b/lib/Importer.js
@@ -148,7 +148,7 @@ export default class Importer {
             this.workingDirectory,
           );
           const results = this.results();
-          results.goto = filePath.startsWith('/')
+          results.goto = path.isAbsolute(filePath)
             ? filePath
             : path.join(this.workingDirectory, filePath);
           resolve(results);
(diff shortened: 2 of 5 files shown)


trotzig:
Thanks! Were you able to test and verify that this change works end-to-end? I know it can be a pain to do that sometimes.

reohjs:
@trotzig This specific change is no longer causing problems but there seems to be [another issue](https://github.com/Galooshi/import-js/issues/399) with sqlite. AFAICT they are not directly related though(?)

reohjs:
I've added the ones that look relevant to absolute paths

lib/normalizePath.js
```diff
@@ -17,7 +18,7 @@ export default function normalizePath(
   }
 
   if (
-    rawPath === '.' || rawPath.startsWith('./') || rawPath.startsWith('../')
+    rawPath === '.' || rawPath.startsWith('./') || rawPath.startsWith('../') || path.isAbsolute(rawPath)
```
trotzig: This doesn't seem quite right. Was this fixing something that was broken before? 🤔

trotzig: I tried this out locally, and it doesn't seem to have any effect. There's a line above that should handle absolute paths (starting with the working directory).

trotzig:
I found another issue that I fixed in https://github.com/Galooshi/import-js/pull/401. Coupled with this change, windows support should be in an okay state again.

lib/normalizePath.js
reohjs: Yeah, I misread that condition. I'll revert.

trotzig: Great! When you have that ready I'll merge and publish a release.

PR #400, opened 2017-05-15
Ignore lowercase jsx identifiers

trotzig:


lib/__tests__/findUndefinedIdentifiers-test.js
```diff
@@ -73,11 +73,21 @@ it('knows about jsx', () => {
...
+    ]),
+  );
+});
+
+it('ignores lowercase jsx element identifiers', () => {
+  expect(findUndefinedIdentifiers(parse("export default <input value='foo' />;"))).toEqual(
+    new Set([
+      'React', // Implicit dependency
+    ]),
+  );
```
trotzig: This is the only significant change to this file (the rest is a result of running prettier on the content).

lib/findUndefinedIdentifiers.js
```diff
@@ -4,20 +4,20 @@ export default function findUndefinedIdentifiers(ast, globalVariables = []) {
...
       }
-      if (!isReference && !definedInScope.has(name)) {
-        result.add(name);
+      if (isReference) {
+        return;
       }
+      if (definedInScope.has(name)) {
+        return;
+      }
+      if (isJSX && /^[a-z]/.test(name)) {
```
lencioni: It might be worth moving the regex up to a constant at the module level, so it doesn't have to be created every time.

trotzig: Good point, that would make it easier to understand what this bit is doing as well.

Issue #351, opened 2016-10-24
Path with backslashes on Windows


moranda:
On windows (at least with atom), path is written with backslashes, which is quite problematic since they are considered as escape characters in the javascript code. They should be automatically replaced with slashes.

trotzig:
I'm being lazy here, as I could try to get access to a windows machine to try things out. But would you be able to try out the latest version of the tool to see if the same issue exists @moranda? Most of the path-handling code in import-js has been rewritten, and it's possible that this issue has been magically fixed somehow.

rndmerle:
Hello.
Still getting backslashes on Windows, using atom-import-js 0.7.0
For instance, I'm getting :

`import { uiSelectors } from '..\Redux\uiRedux';`

trotzig:
Ouch, that's not great. Windows support is something that we've found hard to maintain since the main contributors are all users of either mac or linux. We'll make spikes to fix certain issues (like https://github.com/Galooshi/import-js/pull/401), but then we don't have time to follow up on other changes. Sorry about that.

Fixing this bug is likely not too complicated. A naive solution I could think of is to alter the `pathToImportedModule` variable here, replacing back slashes to forward slashes: 
https://github.com/Galooshi/import-js/blob/a4a284ba5ab710f5fa7eca5b9425214d2d3f813a/lib/JsModule.js#L168

If anyone seeing this thread feels up to the task, I'd be thrilled. I'll provide as much guidance as you need along the way.

rndmerle:
Wait, it's actually my fault.
On my project I've a second package.json further down the tree (in app/) and in fucks everything up, including the slashes. Going to Issue that bug a bit later.

rndmerle:
Wait, it's actually my fault.
On my project I've a second package.json further down the tree (in app/) and it fucks everything up, including the slashes. Going to Issue that bug a bit later.

Slashes are working fine on Windows. Sorry for the troubles.

rndmerle:
Ok actually it's half-good.

`useRelativePaths: false,`
gives
`import { uiSelectors } from 'app/Redux/uiRedux';`

and 
`useRelativePaths: true,`
gives
`import { uiSelectors } from '..\Redux\uiRedux';`

trotzig:
Ah, good catch! That means that it's probably isolated to this this bit of code that uses `path.relative` to find the relative path to use in the import statement: 
[...]

Issue #330, opened 2016-07-31
Use `path.sep` or `path.join` instead of `/`


lencioni:
We haven't done a very good job of using `path.sep` or `path.join` instead of `/` for our paths. We probably should do a better job of this so that things will work better on Windows.

lencioni:
After #413 I'm assuming that we have landed Windows support and no longer need this issue.

PR #401, opened 2017-05-16
Fix indexing node_modules dependencies on Windows

trotzig:
We were assuming a forward slash here, but windows will use backslashes.
This, coupled with #398, will make import-js useful on windows machines
again.


lencioni:
#351 and #330

lencioni:
Lgtm. Are you sure this is the only place we need to do this though?

trotzig:
Coupled with the PR that @reohjs prepared, this was all I needed to do to get importing working again.

trotzig:
Travis failures were the same ones I fixed in #400, so I just ignored them.

PR #402, opened 2017-05-23
Updates ModuleFinder for better symlink tracking.

### LINKED ISSUE #379: ModuleFinder incorrectly assumes symlinked packages live in same-name folder.
`ModuleFinder`’s helper method `assumedLocalPath` makes a bad assumption that symlinked packages live in folders that precisely match their packages. For example, it assumes that a symlinked "mobx" package lives in a folder called precisely `/mobx/`. It's trivial to break this with:

```sh
$ cd /Users/dcporter/Code
$ git clone https://github.com/mobxjs/mobx.git mobx-local
$ cd mobx-local
$ npm link
$ cd ../test-project
$ npm link mobx  # ./node_modules/mobx is now a symlink to a folder called "mobx-local".
```

In this case, `require.resolve` returns `/Users/dcporter/Code/mobx-local/lib/mobx.js`, and `assumedLocalPath` turns that into the nonsensical `"./node_modules//Users/dcporter/Code/mobx-local/lib/mobx.js"`, which super doesn't exist and which fails loudly farther down the line.

The "local checkout with a different name" problem is also likely to be triggered with [scoped packages](https://docs.npmjs.com/misc/scope), whose package names start with "@" and which are nested inside scope folders, e.g. `node_modules/@my-company/package/`; ~~and it's effectively unavoidable if you're using [Lerna](https://github.com/lerna/lerna) to manage a monorepo.~~

A solution is to (recursively) follow the symlink to get the actual target directory, and proceed from there. In the above example, `./node_modules/mobx` points eventually to `/Users/dcporter/Code/mobx-local/`, so that can be used to correctly extract the intra-package part of the path.

I threw together some test code locally and found the following useful:
- `npm link` sets up a two-tiered symlink — one from the linked package into `/usr/local/bin`, and one from there into the project which is using it — so you have to recursively follow the symlink.
- `fs.readlink` works for following one symlink, but I have no honest idea how to do an asynchronous `while` — basically a recursive self-`then`ing Promise — so I cheated and used `fs.readlinkSync` in a `while (true)`.
- In theory, arbitrarily many symlinks can be chained; the way I found to tell when I was done was that `fs.readlinkSync` throws an "EINVAL" error when you give it a non-symlink, so I just `try/catch`’d.
- `npm link` creates an absolute symlink, and `npm link mobx` creates a relative one (🤷🏼‍♂️), so `path.isAbsolute` is handy.
- For relative symlinks, `fs.readlinkSync` appears to return a path relative to the symlink's *parent* folder (e.g. relative to the `node_modules` folder rather than the `node_modules/mobx` folder-cum-symlink), so you need to back up an extra folder when resolving, like: `path.resolve(symlinkLocation, '..', relativeSymlinkTarget);`

dcporter:
Currently it assumes that symlinked packages live in a folder named after the package, which is not always true or practical. Fixes #379.

### DIFF
--- a/lib/ModuleFinder.js
+++ b/lib/ModuleFinder.js
@@ -1,4 +1,5 @@
 import path from 'path';
+import fs from 'fs';
 
 import winston from 'winston';
 
@@ -9,11 +10,6 @@ import findPackageDependencies from './findPackageDependencies';
 import lastUpdate from './lastUpdate';
 import readFile from './readFile';
 import requireResolve from './requireResolve';
-
-function assumedLocalPath(pathToResolvedPackageFile, packageName) {
-  const i = pathToResolvedPackageFile.indexOf(`${path.sep}${packageName}${path.sep}`);
-  return `./node_modules/${pathToResolvedPackageFile.slice(i + 1)}`;
-}
 
 /**
  * Checks for package.json or npm-shrinkwrap.json inside a list of files and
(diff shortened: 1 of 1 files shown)


lib/ModuleFinder.js
```diff
@@ -38,7 +34,56 @@ function expandFiles(files, workingDirectory) {
...
+            // folder*, and we need to resolve it to an absolute path (with the extra '..' making
+            // up for the "link's parent folder" bit).
+            resolvedDirectoryPath = path.resolve(resolvedDirectoryPath, '..', targetPath);
+          }
+        }
+
+        // (Sanity check: make sure that we actually found an ancestor folder of resolvedPath.)
+        if (resolvedPath.indexOf(resolvedDirectoryPath) !== 0) {
+          throw new Error(
+            "expandFiles mismatch: package's resolved path and its local symlink target",
```
trotzig: This message is a little confusing to me. I understand that it's likely to never happen though.

lib/ModuleFinder.js
```diff
@@ -38,7 +34,56 @@ function expandFiles(files, workingDirectory) {
...
+        }
+
+        // With the link's full path, we can strip it from the beginning of resolvedPath.
+        const resolvedPathWithinDirectory = resolvedPath.slice(resolvedDirectoryPath.length);
+
+        // Then tack it on to the package path (pathToResolve), and get it relative to the cwd.
+        const absoluteLocalResolvedPath = path.join(pathToResolve, resolvedPathWithinDirectory);
+        // eslint-disable-next-line prefer-template
+        localPath = './' + path.relative(workingDirectory, absoluteLocalResolvedPath);
+      }
```
trotzig: A few thoughts here: 
- This is quite long. Could you break it out into a function? That would also make it easier to test the thing. 
- Alternatively, we could look at something like https://github.com/kamicane/require-relative and replace this whole block (or perhaps part of it). Because what we're essentially doing here is finding the real path to a package dependency, mimicking the default `require` behavior from node. 
- Alternatively-alternatively, we could just get inspiration from `require-relative` and copy something from its small index.js: https://github.com/kamicane/require-relative/blob/master/index.js

trotzig: I ended up pushing a similar PR here: https://github.com/Galooshi/import-js/pull/404

Would you be able to test that branch out locally and see if it helps? 🙏 If it does, I'll credit the change to you since you found it and made a perfectly valid fix. I just wanted to explore the require-relative path, because I believe it will be more robust. I'm not sure it covers everything that you have here however.

lib/ModuleFinder.js
dcporter: Best I could think of. "expandFiles followed a bunch of symlinks and somehow ended up somewhere else than `require.resolve` waaaat" didn't scan. :)

lib/ModuleFinder.js
dcporter: Confirmed the other branch works (in the narrow `mobx-local` test case at least)! And that PR moves some other stuff in the right direction too (using the file's actual (relative) path as its db key; adding `packageName` all over). I'll shut this one down.

To the implementor goes the implementation credit ;)

dcporter:
Closing in favor of #404.

PR #403, opened 2017-05-23
Teaches findJsModulesFor about scoped packages.

### LINKED ISSUE #380: findJsModulesFor breaks with scoped packages.
`findJsModulesFor` uses a regex to extract the package name from the path ([here](https://github.com/Galooshi/import-js/blob/master/lib/findJsModulesFor.js#L41)) which only matches a single folder name, e.g. `./node_modules/mobx/`. This breaks for [scoped packages](https://docs.npmjs.com/misc/scope), like `@mycompany/package`, which live nested in folders that begin with `@`. I looked through the npm docs for scoped packages and it looks like the leading `@` reliably indicates a scoped package, so this should be a pretty easy fix — if packageName starts with `@`, pivot to a regex which matches the two-folder, @-led package name.

dcporter:
Currently it's unable to find packages like `@company/package`. Fixes #380.

### DIFF
--- a/lib/findJsModulesFor.js
+++ b/lib/findJsModulesFor.js
@@ -26,7 +26,10 @@ function findImportsFromEnvironment(
     );
 }
 
-const PACKAGE_NAME_PATTERN = /\.\/node_modules\/([^/]+)\//;
+// Given a string like './node_modules/package/foo/bar.baz', this is used to
+// extract 'package'. (Also matches scoped packages, for example finding
+// '@company/package' in './node_modules/@company/package/foo/bar.baz'.)
+const PACKAGE_NAME_PATTERN = /\.\/node_modules\/(?:@[^/]+\/)?([^/]+)\//;
 
 function findJsModulesFromModuleFinder(
   config: Configuration,


lib/findJsModulesFor.js
```diff
@@ -27,6 +27,7 @@ function findImportsFromEnvironment(
 }
 
 const PACKAGE_NAME_PATTERN = /\.\/node_modules\/([^/]+)\//;
+const SCOPED_PACKAGE_NAME_PATTERN = /\.\/node_modules\/(@[^/]+\/[^/]+)\//;
```
trotzig: Small nit, but I think the `@` part could be part of the above regexp: 

```javascript
const PACKAGE_NAME_PATTERN = /\.\/node_modules\/(?:@[^/]+\/)?([^/]+)\//;
```
(Note: this is untested and likely to not work directly, because regexps)

dcporter:
(Side note: tests are currently failing on master.)

dcporter:
`findJsModuleFor` [isn't tested for packages](https://github.com/Galooshi/import-js/blob/master/lib/__tests__/findJsModulesFor-test.js#L11), and I'm having trouble writing a non-scoped test to pass on master. I added this to the beforeEach:
```js
      return Promise.resolve([
          { path: './foo.js', isDefault: true },
          { path: './node_modules/pacman/index.js', name: 'pacman', isDefault: true },
      ]);
```
And here's the test:
```js
it('returns matching packages', () => findJsModulesFor(
  new Configuration('./bar.js'),
  'pacman',
).then((jsModules) => {
  expect(jsModules.length).toEqual(1);
  expect(jsModules[0].importPath).toEqual('pacman');
}));
```
Terrifyingly, this is failing with:
```
Expected value to equal: "pacman"
Received: "./foo"
```

lib/findJsModulesFor.js
dcporter: No kidding — it works first pass, nice job.

I'll make the change now.

trotzig:
Thanks for attempting to make the `findJsModulesFor` test better. You're right that we're not testing that path. The test is failing because we're mocking out `ModuleFinder.find` to always return the same thing. I think you can make it work by changing the mock to something like: 

```js
  ModuleFinder.getForWorkingDirectory.mockImplementation(() => ({
    find(variableName) {
      if (variableName === 'foo') {
        return Promise.resolve([{ path: './foo.js', isDefault: true }]);
      }
      if (variableName === 'pacman') {
        return Promise.resolve([{ path: './node_modules/pacman/index.js', name: 'pacman', isDefault: true, packageName: 'pacman' }]);
      }
    },
  }));
```

(That last `packageName` because of changes I made in #404). 

That being said, the original fix in this commit will be a little hard to apply on top of 1390c77. I wonder if this change is even necessary after that change? Give me a second and I'll test. :)

dcporter:
(Testing by hand, but could you release version 2.2.3 so the plugins catch up?)

dcporter:
Looks good! Closing out as already fixed by 1390c77.

Official request for v2.2.3!

PR #412, opened 2017-06-05
Find nested exports when inside ternary

### LINKED ISSUE #411: Named exports detection foiled by top-level conditional
Some packages, like airbnb-prop-types, will have different builds for production vs development, in order to minimize bundle size. In this case, the entry point looks something like this:

```js
module.exports = process.env.NODE_ENV === 'production'
  ? require('./build/mocks')
  : require('./build');
```

https://unpkg.com/airbnb-prop-types@2.5.4

This ends up breaking the named exports detection for this package. I'm not entirely sure what the best solution is here, but in this case it would work if import-js followed either path, since they should both be exporting the same set of things. Perhaps we could add some special cases for `process.env.NODE_ENV` checks and follow the production path?

cc @dabbott

trotzig:
Some libraries (like [airbnb-prop-types](https://unpkg.com/airbnb-prop-types@2.5.4)), like to export different
files depending the current NODE_ENV. Export statements can then look
something like:

```js
  module.exports = process.env.NODE_ENV === 'production' ?
    require('a') : require('b');
```

In most cases, the two alternatives will have the same structure, so
finding named exports in either of the two should be fine.

I've made it so that we always pick the first alternative (the
`consequent`) in ternary exports.

There are of course other ways of conditionally exporting things. For
now I'm only dealing with ternaries, we can worry about other styles
when we have those use-cases.

Fixes #411

### DIFF
--- a/lib/__tests__/findExports-test.js
+++ b/lib/__tests__/findExports-test.js
@@ -259,26 +259,6 @@ it('finds CommonJS exports in a root, self-executing, function', () => {
   });
 });
 
-it('follows exported requires', () => {
-  fs.__setFile(
-    '/path/to/foo.js',
-    'const Result = { bar: 123 }; module.exports = Result;',
-    { isDirectory: () => false },
-  );
-  requireRelative.resolve.mockImplementation(() => '/path/to/foo.js');
-  expect(
-    findExports(
-      `
-    module.exports = require('./foo');
-  `,
-      '/path/to/file.js',
-    ),
-  ).toEqual({
-    named: ['bar'],
-    hasDefault: true,
-  });
-});
-
 it('finds inner CommonJS exports', () => {
   // this is from logpath
   expect(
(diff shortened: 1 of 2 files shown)


lib/__tests__/findExports-test.js
```diff
@@ -459,3 +439,38 @@ it('does not fail for inner exports that are not objects', () => {
...
+      hasDefault: true,
+    });
+  });
+
+  it('picks the first require if inside a ternary', () => {
+    expect(
+      findExports(
+      `
+        module.exports = process.env.NODE_ENV === 'test' ?
+          require('./foo') : require('./ignored');
```
lencioni: What if the first branch of the ternary doesn't have a require but the second branch does?

lib/findExports.js
```diff
@@ -37,6 +37,31 @@ function findESNamedExports(node) {
...
+  if (node.type === 'ConditionalExpression') {
+    // Potential ternary-style export - we pick the first one
+    // module.exports = foo ? require('a') : require('b');
+    return resolveNestedNamedExports(node.consequent, absolutePathToFile);
+  }
+  if (
+    node.type === 'CallExpression' &&
+    node.callee.name === 'require' &&
+    node.arguments.length === 1 &&
+    node.arguments[0].type === 'StringLiteral'
```
lencioni: unrelated: I wonder if conditionals like this would be clearer if they were wrapped up in well-named functions. Here something like `isNodeRequireStatement`?

lib/__tests__/findExports-test.js
trotzig: Yeah, it will just stop trying to find exports at that time. I'd assume this is unlikely though, so we can probably roll with this until we have a counterexample.

lib/findExports.js
trotzig: Yeah, there are definitely some room for improvements here. I find it a little hard to name though in some cases. The names in the AST are pretty abstract already, and in some cases we look for even more abstract constructs. I'll remember this for next time in here though.

PR #413, opened 2017-06-05
Taking care of Windows' file pathes

rndmerle:
Ok, tests pass both on Windows and Linux.
I can adapt names and stuff as needed.

### DIFF
--- a/lib/WindowsCompat.js
+++ b/lib/WindowsCompat.js
@@ -1,5 +1,0 @@
-export default {
-  pathToCanonical(path: string): string {
-    return path.replace(/(^[A-Z]:\\|\\)/g, '/');
-  },
-};
--- a/lib/WindowsCompat.js
+++ b/lib/WindowsCompat.js
@@ -1,0 +1,5 @@
+export default {
+  pathToCanonical(path: string): string {
+    return path.replace(/(^[A-Z]:\\|\\)/g, '/');
+  },
+};
--- a/lib/findProjectRoot.js
+++ b/lib/findProjectRoot.js
@@ -2,7 +2,7 @@ import fs from 'fs';
 import path from 'path';
 
 function findRecursive(directory) {
-  if (directory === '/') {
+  if (directory === '/' || /^[A-Z]:\\$/.test(directory)) {
     throw new Error('No project root found, looking for a directory with a package.json file.');
   }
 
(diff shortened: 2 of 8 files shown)


lib/WindowsCompat.js
```diff
@@ -0,0 +1,5 @@
+export default {
+  pathToCanonical(path: string): string {
+    return path.replace(/(^[A-Z]:\\|\\)/g, '/');
```
trotzig: Since backslashes are allowed in *nix filenames, I think we should guard here against replacing so that it only happens on windows. Would that make sense?

lib/findProjectRoot.js
```diff
@@ -2,7 +2,7 @@ import fs from 'fs';
 import path from 'path';
 
 function findRecursive(directory) {
-  if (directory === '/') {
+  if (directory === '/' || /^[A-Z]:\\$/.test(directory)) {
```
trotzig: For marginally better performance, you could initialize this regex in a constant at the top of the file.

lib/WindowsCompat.js
```diff
@@ -0,0 +1,5 @@
+export default {
```
trotzig: Instead of exporting an object with a method, you could just export a function directly. Maybe call it `forwardSlashes` or something to make obvious what it's doing.

lib/WindowsCompat.js
rndmerle: If we get a backslash in path I guess it must be escaped right?
So we'd just need to avoid replacing double-backslashes?

lib/WindowsCompat.js
rndmerle: Totally fine. I've no idea why I chose this way at the time.

lib/findProjectRoot.js
rndmerle: And better code readability actually :)

lib/WindowsCompat.js
rndmerle: Problably in case more compat functions get added. But exporting named functions cuts it too.

trotzig:
I plan on doing some manual testing tomorrow, and then I'll ship this in a new release. Thanks again for the contribution!

rndmerle:
I'm glad I could help.

Not sure you noticed but I didn't change the regex in `forwardSlashes()` yet (about the possible antislash in nix pathes), I was waiting for more infos. But you can proceed as you see during your merge.

trotzig:
Yeah, I saw. I think we can wait for a bug report before we adress that. It might be a non-issue.

PR #414, opened 2017-06-05
Add full text search

billyvg:
I would like to have this feature so that autocomplete plugins (i.e. deoplete) can be supported. The workflow would be more streamlined this way.

Edit:
Here is the [deoplete plugin](https://github.com/billyvg/deoplete-import-js)

### DIFF
--- a/lib/ExportsStorage.js
+++ b/lib/ExportsStorage.js
@@ -226,14 +226,9 @@ export default class ExportsStorage {
         const promises = names.map(name =>
           this._insert({ name, pathToFile, isDefault: false, packageName }));
         promises.push(
-          ...defaultNames.map((defaultName: string | Object) => {
-            const defaultNameObj = typeof defaultName === 'string' ? { name: defaultName, additional: false } : defaultName;
-            const {
-              name,
-              additional,
-            } = defaultNameObj;
-            return this._insert({ name, pathToFile, isDefault: true, packageName, additional });
-          }),
+          ...defaultNames.map(({ name, additional }) =>
+            this._insert({ name, pathToFile, isDefault: true, packageName, additional }),
+          ),
         );
         return Promise.all(promises);
       }));
--- a/lib/ExportsStorage.js
+++ b/lib/ExportsStorage.js
@@ -52,6 +52,17 @@ export default class ExportsStorage {
 
             this.db.run(
               `
+                CREATE VIRTUAL TABLE exports_fts USING fts4(
+                  name,
+                  path,
+                  isDefault,
+                  packageName,
+                )
+              `,
+            );
+
+            this.db.run(
+              `
                 CREATE TABLE mtimes (
                   path TEXT,
                   mtime NUMERIC
@@ -178,7 +189,7 @@ export default class ExportsStorage {
     });
   }
 
-  _insert({ name, pathToFile, isDefault, packageName }) {
+  _insert({ name, pathToFile, isDefault, packageName, additional }) {
     const exportName = isDefault ? normalizedExportName(name) : name;
     return new Promise((resolve, reject) => {
       this.db.run(
@@ -192,6 +203,17 @@ export default class ExportsStorage {
             reject(err);
             return;
           }
+
+          if (!additional) {
+            this.db.run(
+              'INSERT INTO exports_fts VALUES (?, ?, ?, ?)',
+              name,
+              pathToFile,
+              isDefault,
+              packageName,
+            );
+          }
+
           resolve();
         },
       );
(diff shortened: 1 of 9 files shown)


lib/ExportsStorage.js
```diff
@@ -192,6 +203,17 @@ export default class ExportsStorage {
             reject(err);
             return;
           }
+
+          if (!additional) {
```
trotzig: I had to read the code a few times to understand what this boolean meant. A comment here could probably make things clearer. 🤔

lib/ExportsStorage.js
```diff
@@ -52,6 +52,17 @@ export default class ExportsStorage {
 
             this.db.run(
               `
+                CREATE VIRTUAL TABLE exports_fts USING fts4(
```
trotzig: Can you elaborate a little why the virtual table is necessary here? I could see this being implemented using "simple" `WHERE name LIKE 'foobar%'` as well. The redundancy is probably okay, I'm asking more because I'm (passively) searching for alternatives to sqlite3. Multiple people have reported that it's hard to install (e.g. https://github.com/Galooshi/atom-import-js/issues/15, https://github.com/Galooshi/atom-import-js/issues/18) and it would be easier if there was a "pure" javascript alternative.

lib/ExportsStorage.js
```diff
@@ -192,6 +203,17 @@ export default class ExportsStorage {
             reject(err);
             return;
           }
+
+          if (!additional) {
+            this.db.run(
+              'INSERT INTO exports_fts VALUES (?, ?, ?, ?)',
```
trotzig: `fts` is for full text search, right? I think it would be okay to spell that out instead of using the abbreviation.

lib/ExportsStorage.js
```diff
@@ -296,4 +329,31 @@ export default class ExportsStorage {
...
+          COLLATE NOCASE
+        `,
+        variableName,
+        (err, rows) => {
+          if (err) {
+            reject(err);
+            return;
+          }
+          resolve(
+            rows.map(({ name, path, isDefault, packageName }) => ({
```
trotzig: If you wanted to DRY up things, you could introduce a normalization function (`normalizeRowData`?) and use it here and inside the `get()` method.

lib/ExportsStorage.js
```diff
@@ -204,8 +226,14 @@ export default class ExportsStorage {
         const promises = names.map(name =>
           this._insert({ name, pathToFile, isDefault: false, packageName }));
         promises.push(
-          ...defaultNames.map(name =>
-            this._insert({ name, pathToFile, isDefault: true, packageName })),
+          ...defaultNames.map((defaultName: string | Object) => {
+            const defaultNameObj = typeof defaultName === 'string' ? { name: defaultName, additional: false } : defaultName;
```
trotzig: This line looks a little defensive. Can we make sure that we only pass in objects instead?

lib/__tests__/findJsModulesFor-test.js
```diff
@@ -19,7 +19,6 @@ beforeEach(() => {
 it('returns matching modules', () => findJsModulesFor(
   new Configuration('./bar.js'),
   'foo',
-  './bar.js',
```
trotzig: Good catch!

lib/ExportsStorage.js
billyvg: sqlite's full text search requires a virtual table and using full text search performs better than `LIKE 'foo%'`. I'm not too familiar with sqlite, but it's possible we don't need a separate table at all, but I figured this way I don't break anything.

trotzig:
Ready to go? 

One thing that crossed my mind when looking at the deoplete plugin is that, since you're using the command-line tool directly, there's a dependency on a warm cache. If people use your plugin without also using vim-import-js (actively), the items they need won't be in import-js storage. If you wanted to fix that you could use a daemon `importjs` process to communicate. Or better yet, reuse the daemon process from vim-import-js. That way you'd make sure that the cache is in a better state when you autocomplete.

billyvg:
Should be good to go now!

For the deoplete plugin, I should probably call `importjs#ExecCommand` instead, yeah? (And also add a search function in `vim-import-js`)

lib/ExportsStorage.js
```diff
@@ -8,6 +8,14 @@ function normalizedExportName(string) {
   return string.toLowerCase().replace(/[-_.]/g, '');
 }
 
+function normalizeRows(rows) {
+  // currently just casts `isDefault` to boolean
+  return rows.map(({ isDefault, ...other }) => ({
+    isDefault: !!isDefault,
+    ...other,
```
trotzig: nice!

trotzig:
Yeah, that sounds like the right thing. You'd also have to add search to the list of commands in `lib/daemon.js`. That's the entry point for all interaction between vim-import-js and import-js.

trotzig:
I plan on releasing this tomorrow after some manual testing locally. Thanks again for the contribution @billyvg!

lencioni:
This is perfect timing. I was just chatting with @dabbott about using import-js with autocomplete. It would be great if it added the import after selecting the item from the autocomplete list.

trotzig:
I'm very excited about this as well. If I read the source for the deoplete plugin right, I believe it did import the variable as well.

billyvg:
Yep the deoplete plugin will add the import statement - though there are some small annoying bugs with it.

PR #417, opened 2017-06-07
Search through daemon

trotzig:
Follow-ups on the support for full-text search that was added in #414.


trotzig:
Not sure why Node 4 is so slow (some tests are even timing out). Will merge anyway.

Issue #418, opened 2017-06-08
Default parameters shouldn't import


dabbott:
Currently when using an ES2015 method definition with a default parameter, e.g:

```js
const obj = {
  method(foo = 1) {}
}
```

`foo` will be imported. This only happens when there's a default parameter. It also happens for setters (though probably this would never happen in real code):

```js
const obj = {
  set value(foo = 1) {}
}
```

lencioni:
Good find! I'm guessing this is a bug in `findUndefinedIdentifiers`. Would you be interested in opening a PR? (Even a PR with just a failing test would be awesome!)

dabbott:
Fixed! It turned out that default parameters in general don't get imported.

PR #420, opened 2017-06-08
Recognize default parameters

dabbott:
Recognizes default parameter assignment (fixes https://github.com/Galooshi/import-js/issues/418)

### DIFF
--- a/lib/__tests__/findUndefinedIdentifiers-test.js
+++ b/lib/__tests__/findUndefinedIdentifiers-test.js
@@ -178,6 +178,22 @@ it('knows about objects', () => {
   ).toEqual(new Set(['foo', 'uuid']));
 });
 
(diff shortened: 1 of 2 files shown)


lib/__tests__/findUndefinedIdentifiers-test.js
```diff
@@ -178,6 +178,22 @@ it('knows about objects', () => {
...
+        `
+    foo({
+      bar(a = 123) {},
+    });
+
+    function x(b = 456) {}
+  `,
+      ),
+    ),
+  ).toEqual(new Set(['foo']));
```
trotzig: Sorry for the weird formatting used in this file, and thanks for rolling with it. I believe it's a result of running `prettier` on the code.

trotzig:
Thanks @dabbott!

Issue #422, opened 2017-06-09
Including import statement in unresolved imports resolution dialog


dabbott:
Currently the unresolved imports dialog can be a little confusing. Showing file paths tends to create a lot of noise. We could potentially leave off the file path, but there isn't quite enough information to make a good choice from just the `displayName`

E.g:
![screen shot 2017-06-09 at 11 28 33 am](https://user-images.githubusercontent.com/1198882/26989186-fe290f28-4d06-11e7-9bfe-9f3234e4dff8.png)

It would be easier to make the right choice if we showed the actual import statement. The choices are pretty clear when displayed like this:

![screen shot 2017-06-09 at 11 15 24 am](https://user-images.githubusercontent.com/1198882/26989237-3367dfc0-4d07-11e7-8d54-e7c585fa1d92.png)

In cases where two import statements look exactly the same, the full path will still need to be shown -- but it can probably be left to the editor plugin to figure out the best way to display things.

lencioni:
Closed by #423

PR #423, opened 2017-06-09
Include import statement in list of unresolved imports

dabbott:
Adds support for #422

Presumably there will always be exactly one import string for the unresolved imports use case, so it's accurate to access it as [0]. And there will never be a multiline import string, so the tab string isn't used. But happy to change the implementation.

### DIFF
--- a/lib/Importer.js
+++ b/lib/Importer.js
@@ -410,6 +410,9 @@ export default class Importer {
         this.pathToCurrentFile,
         this.workingDirectory,
       ),
+      importStatement: jsModule
+        .toImportStatement(this.config)
+        .toImportStrings(Infinity, '  ')[0],
     }));
 
     return undefined;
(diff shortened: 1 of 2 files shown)


lib/Importer.js
```diff
@@ -410,6 +410,9 @@ export default class Importer {
         this.pathToCurrentFile,
         this.workingDirectory,
       ),
+      importStatement: jsModule
```
trotzig: Instead of adding another key, I think you should just change `displayName` to be the import statement.

dabbott: Good thinking!

dabbott:
@trotzig updated!

lib/Importer.js
```diff
@@ -404,7 +404,9 @@ export default class Importer {
     this.unresolvedImports[variableName] = jsModules.map((
       jsModule: JsModule,
     ): Object => ({
-      displayName: jsModule.displayName(),
```
lencioni: I wonder if we still need this method on jsModule.

dabbott: I think in some contexts it's a little nicer than showing the import statement https://github.com/Galooshi/import-js/blob/master/lib/Importer.js#L98 

¯\_(ツ)_/¯

PR #425, opened 2017-06-09
Sort unresolved imports by path distance to the current file

dabbott:
Uses the simple heuristic of path distance to rank imports.

Closer paths are frequently the right ones: probably from `./a.js` you're looking for `./foo.js` rather than `./something/very/deeply/nested/foo.js` or `../../../../foo.js`.

node_modules will have an `importPath` like `react`, so they will generally come before deeply nested imports (though there could be cases where they don't).

I'm sure there are better ranking algorithms but this one is pretty straightforward works reasonably well for me 😄

### DIFF
--- a/lib/Importer.js
+++ b/lib/Importer.js
@@ -401,6 +401,11 @@ export default class Importer {
       return undefined;
     }
 
+    const distanceToCurrentFile = (importPath: string): number => {
+      const match = path.relative(this.pathToCurrentFile, importPath).match(/\//g);
+      return match ? match.length : 0;
+    };
+
     this.unresolvedImports[variableName] = jsModules.map((
       jsModule: JsModule,
     ): Object => ({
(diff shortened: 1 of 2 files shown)


lib/Importer.js
```diff
@@ -401,6 +401,11 @@ export default class Importer {
       return undefined;
     }
 
+    const distanceToCurrentFile = (importPath: string): number => {
+      const match = path.relative(this.pathToCurrentFile, importPath).match(/\//g);
```
trotzig: The `path.relative` here is most often redundant, right? Unless `useRelativePaths` is `false`.

dabbott: That sounds right... should we check for that?

Now that I think about it, I should probably check for path.sep instead of / to handle the windows case? Or are these paths normalized already?

trotzig: These paths should be normalized, so no need to use `path.sep`. I'll let you make the call on the `path.relative`. Either just remove it (and not worry about relative or absolute import paths), or keep it around for slightly better absolute path support. Let me know when you're done and I'll merge.

dabbott:
@trotzig updated to remove `path.relative`

trotzig:
Thanks! And sorry for the delay, I'm in vacation mode. 🌴

I'll try to release something tomorrow.

PR #427, opened 2017-06-13
Moved checking for excludes to allow excluding node_modules

dabbott:
We can now exclude specific node_modules using `excludes`

### DIFF
--- a/lib/findJsModulesFor.js
+++ b/lib/findJsModulesFor.js
@@ -50,6 +50,12 @@ function findJsModulesFromModuleFinder(
         const modules = exports.map((
           { name, path, isDefault, packageName }: Object,
         ): ?JsModule => {
+          // Filter out modules that are in the `excludes` config.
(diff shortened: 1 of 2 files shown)


lib/findJsModulesFor.js
```diff
@@ -50,6 +50,12 @@ function findJsModulesFromModuleFinder(
         const modules = exports.map((
           { name, path, isDefault, packageName }: Object,
         ): ?JsModule => {
+          // Filter out modules that are in the `excludes` config.
+          if (config.get('excludes').some((glob: string): boolean =>
```
lencioni: nit: I think this would be more readable if you used a variable here. something like this:

```js
const isExcluded = config.get('excludes')
  .some((glob: string): boolean => minimatch(path, glob));
if (isExcluded) {
  return undefined;
}
```

It looks like you are just moving this code from somewhere else!

Issue #284, opened 2016-05-31
Provide option to disable import re-ordering


kesne:
Import re-ordering can cause a lot of git diff churn, especially with larger teams. Ideally, this would be a one-time churn, but not everybody on large teams may be using import-js to create every single import, and as such it's easy for new imports to not end up alpha-sorted, causing further automatic imports to have churn.

Ideally, the sorting of import could be disabled to prevent this issue, and simply have new imports appended.

trotzig:
Hi @kesne!

I understand that sorting (and grouping) can be annoying in a large team where everyone isn't using import-js. Adding such a configuration option should be fairly simple (we can look at how [`groupImports` controls grouping in `ImportStatements](https://github.com/Galooshi/import-js/blob/004b0ad01b7a5ab9987be71bf512b33ec2495b32/lib/ImportStatements.js#L219) for inspiration). 

An alternative solution for your team could be to turn import-js into a linter (maybe an eslint plugin?). Then people could use import-js and pass the linter, or do it manually and have to fix lints. The end result would be consistently sorted imports, without forcing people to use import-js.

kesne:
For sure, I'll take a look at that and hopefully come up with something.

Personally, I don't care enough about sorted imports (or unsorted imports) to enforce it via a lint rule.

IanVS:
I would also like the option to turn off re-ordering, especially as the grouping done by this tool is different from the order that I am already enforcing via [`eslint-plugin-import/order`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md),

trotzig:
@IanVS Instead of turning off ordering, would it help if import-js supported the same config as that eslint plugin? `groups` and `newlines-between`. 

I'm also interested to hear if @lencioni has any input here. Turning off reordering would probably mean we just stick the import at the top. Would that be sufficient?

IanVS:
Yeah, I mean ideally if it supported those same options and sorted based on them it would be great, but I could foresee that getting fairly complicated.  As a first step, disabling sorting and throwing the new import at the top would be totally sufficient for me personally.  I can then just move that one line into place, and not worry about all the other lines being shuffled around too.

trotzig:
[...]

PR #432, opened 2017-06-19
Add sortImports option to enable/disable import sorting

dabbott:
I've added a config option to disable sorting, per https://github.com/Galooshi/import-js/issues/284


dabbott:
I'm in favor of merging this as-is, and then figuring out what the function version would look like separately.

Probably the `toNormalized` version of the `ImportStatement` isn't really what we'd want to pass to the comparator (not enough flexibility). Possibly we pass the whole `ImportStatement`? Or maybe the `toImportStrings` version? Or...?

trotzig:
I'll add something to the readme in a bit.

trotzig:
On second thought, given the plan in #433, I'm going to leave this undocumented for now. Let's add it once we've settled on something.

Issue #215, opened 2016-03-03
Comments in import blocks


lencioni:
I've encountered some code that had the import block divided up with comments already. Example:

```js
import Foo from 'foo';
import Bar from 'bar';

// Fruits
import Banana from '../banana';
import Orange from '../orange';

// Animals
import Cat from '../cat';
import Dog from '../dog';
```

This seems okay to me, but import-js stopped looking for imports once it hit the first comment, which isn't great. I'm not entirely sure what the right thing is to do here, but it seems like what we currently have isn't great. I suppose we could disable grouping in cases like this and then put the import somewhere reasonable--maybe a separate block at the end?

trotzig:
Yeah, not sure what to do here. Identifying this pattern and putting the imports in a separate block sounds complicated. I think this might be one case where you just have to adapt the import style so that it's consistent with other files. :dragon_face:

rhettlivingston:
I do not believe that the example shown above is one that we can truly handle because we could never truly determine that a comment applies to multiple lines. We could make a guess based on the combination of the blank line and the comment, but then what would we do? Keep the comment and the whole block together? Auto-disable grouping?

However, I was surprised in another case to discover that comments stopped the scan for imports. My expectation based on automatic code beautifiers I've used in the past (which often had a feature of sorting and grouping includes) was that the comments would be kept and passed through with the statement that they apparently applied to. The assumption is usually made that a comment applies to the statement after it except when the statement appears on the same line.

So I believe that
```js
import hijk from 'localModule'; // trailing comment for local module.
/*
* package leading comment
*/
import cdef from 'package'; // package trailing comment
// CoreModule leading comment
import abcd from 'coreModule';
```
should become
```js
// CoreModule leading comment
import abcd from 'coreModule';

/*
* package leading comment
*/
import cdef from 'package'; // package trailing comment

import hijk from 'localModule'; // trailing comment for local module.
```
assuming grouping is enabled. 

If grouping is disabled, they should still be tolerated and passed through with the apparently associated package.

The exception would be that comments at the top of the file must remain at the top of the file. They are very possibly instructions for tools like lint. Perhaps an exception to the exception could be made if blank lines were found before the comment and there are no blank lines between the comment and the first import.

lencioni:
> The assumption is usually made that a comment applies to the statement after it except when the statement appears on the same line.

This is usually true, except for cases where you do something like disable a linting rule with a comment, and then re-enable it. e.g.

```js
foo = bar;

[...]

PR #433, opened 2017-06-20
Preserve comments and whitespace

dabbott:
This is a rough first pass. 

I want to preserve comments (https://github.com/Galooshi/import-js/issues/215) so that we can `fix` imports without losing things like eslint line rules, manually added labels for groups of imports, etc. I think it would be hard to perfectly capture every use case with the same setting, so there might need to be multiple settings. 

By far the most common kind of comment/whitespace within imports in our codebase is a comment directly before an import, optionally with a newline above it:

```
import foo from 'foo'

// Comment
import bar from 'bar'
```

The behavior I chose is: an import can have "leading content", which will follow the import if it moves. In this example, the newline and comment would follow `import bar` if it's moved. If sorting is disabled, this means comments and whitespace will remain exactly as they are, with new imports added on top.

Comments before the imports are untouched, since they might be large block comments (license) or other things we don't want to move. Comments after the imports are also untouched.

(This is on top of my sortImports changes)

### DIFF
--- a/lib/ImportStatement.js
+++ b/lib/ImportStatement.js
@@ -35,6 +35,7 @@ export default class ImportStatement {
   importFunction: ?string;
   namedImports: Array<Object>;
   originalImportString: ?string;
+  leadingContent: ?string;
   path: string;
 
   constructor(
@@ -46,6 +47,7 @@ export default class ImportStatement {
       importFunction,
       namedImports = [],
       originalImportString,
+      leadingContent,
       path,
     }: {
       assignment?: ?string,
@@ -55,6 +57,7 @@ export default class ImportStatement {
       importFunction?: ?string,
       namedImports?: Array<Object>,
       originalImportString?: ?string,
+      leadingContent?: ?string,
       path: string,
     } = {},
   ) {
@@ -65,6 +68,7 @@ export default class ImportStatement {
     this.importFunction = importFunction;
     this.namedImports = namedImports;
     this.originalImportString = originalImportString;
+    this.leadingContent = leadingContent;
     this.path = path;
   }
 
@@ -164,6 +168,16 @@ export default class ImportStatement {
   }
 
   toImportStrings(maxLineLength: number, tab: string): Array<string> {
+    const strings = this._importStrings(maxLineLength, tab);
+
+    if (this.leadingContent && strings.length > 0) {
+      strings[0] = this.leadingContent + strings[0];
+    }
+
+    return strings;
+  }
+
+  _importStrings(maxLineLength: number, tab: string): Array<string> {
     if (this.originalImportString) {
       return [this.originalImportString];
     }
--- a/lib/__tests__/Importer-test.js
+++ b/lib/__tests__/Importer-test.js
@@ -512,6 +512,159 @@ foo
         },
       );
 
+      describe('when sorting and grouping are disabled', () => {
+        beforeEach(() => {
+          configuration.groupImports = false;
+          configuration.sortImports = false;
+        });
+
+        it('leaves newlines', () => {
+          text = `
+import bar from './bar';
+
+import baz from './bar/baz';
+
+foo
+          `.trim();
+
+          expect(subject()).toEqual(
+            `
+import foo from './bar/foo';
+import bar from './bar';
+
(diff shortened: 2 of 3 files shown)


lib/Configuration.js
```diff
@@ -64,6 +65,7 @@ const DEFAULT_CONFIG = {
   stripFileExtensions: ['.js', '.jsx'],
   tab: '  ',
   useRelativePaths: true,
+  whitespace: 'strip',
```
lencioni: I'm a little hesitant to introduce a bunch of extra config options. Is there any compelling reason to not just do this for everyone?

dabbott:
About introducing settings:

For **whitespace**, I'd imagine the desired behavior of import-js might differ per project. I might:
1. Prefer cleaning up imports with sorting/grouping/whitespace stripping
2. Prefer changing imports as little as possible, for less noisy diffs on a large project

Rather than introducing another config option, it could make sense to tie this together with `sortImports` or `groupImports`. If I prefer (1), then I want sorting and misc whitespace stripped. If I prefer (2), then I probably set `sortImports: false, groupImports: false` and want to leave whitespace as-is.

It doesn't really make sense to try to preserve whitespace if we also sort imports, since the whitespace will move around, seemingly at random. And it doesn't really make sense to try to preserve whitespace if we also group imports, since grouping inherently modifies whitespace -- with the exception: if whitespace is above a non-eslint comment, I may be attempting to group and label my imports manually (I'd say this is more of a comments-related use case though).

---

For **comments**, I think they *should* be preserved for everyone. However, the implementation has to be much smarter when it's one-size-fits-all and can't be disabled. For this case:

```js
/* eslint-disable some-rule */
import a from 'a';
import b from 'b';
/* eslint-enable some-rule */
```

we would need to recognize that these imports and comments must be moved together. Similarly for:

```js
import Foo from 'foo';
import Bar from 'bar';

// Fruits
import Banana from '../banana';
import Orange from '../orange';
```

we would have to decide if these are recognized as 2 groups to keep together, or are just coincidentally grouped. If we move (via sorting/grouping) Orange into another position, the comment is no longer useful. However we probably shouldn't try to keep these together:

```js
import Foo from 'foo';
import Bar from 'bar';

// eslint-disable-next-line some-rule
import Banana from '../banana';
import Orange from '../orange';
```

Or we keep groups of imports together more aggressively than necessary, but this may start to seem like we're grouping at random.

We could allow a function config option, where the intent of the comment could be specified:

```js
{
  commentAnalyzer: ({ text, multiline, hasLeadingBlankLine }) => {
    if (/eslint-disable-next-line/.test(text.value) return {leading: true};
    if (text.multiline && /eslint-enable/.test(text.value) return {leading: true, group: true};
    if (text.multiline && /eslint-disable/.test(text.value) return {trailing: true, group: true};
    if (text.hasLeadingBlankLine) return {leading: true, group: true};
    return {leading: true};
  }
}

And then we could use these hints to do the right thing. The default for this config option could be whatever seems most common, but it could be overridden when we get it wrong.

Another simpler approach: if I have `sortImports: false, groupImports: false` I want comments to be left alone. We could just handle this case and recommend turning off sorting and grouping if you need comments untouched.
```

---

Other misc idea:

We could add an option that captures the intent of the person who wants to minimize noisy diffs. No sorting, grouping, and comments/whitespace untouched.

lib/Configuration.js
dabbott: Hmm... not sure. I left my thoughts in the main PR comments

dabbott:
@trotzig @lencioni what do you guys think is the best way to handle whitespace and comments?

trotzig:
I'm in favor of the simple solution to begin with, looking at the combination of `sortImports` and `groupImports`. If both of those are false, leave comments and whitespace alone.

dabbott:
That sounds good to me. It handles my use case for sure. With this in, we'll start using import-js for the Airbnb React Native codebase :D

Updated PR to look at `sortImports` and `groupImports` instead of adding new config options @lencioni

lencioni:
Thanks! I'm going to merge this. Can you follow up with a PR that adds documentation for `sortImports` (#432)? Once that is in, I'll publish a new release.

PR #435, opened 2017-06-21
Add support for `export * from`, export as default and more obtuse ES6 export syntaxes

coagmano:
Added support for:
- Recursive export from file `export * from './foo'`
- namespace exports `export * as ns from './foo'`
  (uses babylon `exportExtensions` plugin)
- Renaming as default `export { foo, bar as default }`
- Renaming during destructure `export const { foo, bar: baz } = someObject`
- Default values `export const { foo, bar = {a:1, b:2} } = someObject`
- Rename during destructure with defaults:
    `export const {
      foo,
      bar: baz = {a: 1, b: 2}
    } = someObject`
- Rest parameter `export const { foo, ...bar } = someObject`
- Array destructure `export const [ foo, bar ] = someObject`
- Array rest parameter `export const [ foo, ...bar ] = someObject`

### DIFF
--- a/lib/findExports.js
+++ b/lib/findExports.js
@@ -5,7 +5,18 @@ import requireRelative from 'require-relative';
 
 import parse from './parse';
 
-function findESNamedExports(node) {
+function findESNamedExports(
+  node,
+  {
+    absolutePathToFile,
+  },
+) {
+  if (node.type === 'ExportAllDeclaration') {
+    // Recurse the file referenced by the declaration
+    // eslint-disable-next-line no-use-before-define
+    return resolveNestedNamedExports(node, absolutePathToFile);
+  }
+
   if (node.type !== 'ExportNamedDeclaration') {
     return [];
   }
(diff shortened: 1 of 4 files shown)


lib/findExports.js
```diff
@@ -27,11 +38,49 @@ function findESNamedExports(node) {
...
-    } else {
+    // Always check for type === 'Identifier' before adding name
+    if (id.type === 'Identifier') {
+      // export const foo = 'bar'
       result.push(id.name);
+    } else if (id.type === 'ObjectPattern') {
+      // export const { foo, bar } = something();
+      id.properties.forEach((property) => {
+        if (property.type === 'ObjectProperty') {
+          // use value instead of key, since destructuring can remapped the
```
trotzig: did you mean "can remap"?

coagmano: Ha, yes I did.

trotzig: No worries. e0a15a1

PR #436, opened 2017-06-21
Update Meteor environment to support api.mainModule

coagmano:
Update to `extractExportsFromMeteorPackage` in `meteorEnvironment.js` to look
for a mainModule in the isoPack and uses the updated findExports from #435 to
parse the source file for named exports.

I could use some help adding tests for this. 
Should I import meteorEnvironment, mock an isoPack file and feed that into 
the `namedExports` function?

### DIFF
--- a/lib/__tests__/Configuration-test.js
+++ b/lib/__tests__/Configuration-test.js
@@ -2,6 +2,7 @@
 import fs from 'fs';
 import os from 'os';
 import path from 'path';
+import requireRelative from 'require-relative';
 
 import Configuration from '../Configuration';
 import FileUtils from '../FileUtils';
@@ -10,6 +11,7 @@ import version from '../version';
 jest.mock('../FileUtils');
 jest.mock('../version');
 jest.mock('fs');
+jest.mock('require-relative');
 
 describe('Configuration', () => {
   afterEach(() => {
@@ -381,6 +383,8 @@ describe('Configuration', () => {
         `
 john:foo
 jane:bar
+john:foobar
+jane:barbaz
         `.trim(),
         { isDirectory: () => false },
       );
@@ -389,6 +393,8 @@ jane:bar
         `
 john:foo@1.0.0
 jane:bar@0.0.1
+john:foobar@1.0.0
+jane:barbaz@0.0.1
         `.trim(),
         { isDirectory: () => false },
       );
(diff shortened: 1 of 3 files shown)


coagmano:
Would also love help getting it to provide a correct path for local atmosphere packages. Even with these changes it still gives me the choice of these:
```
import { Utils } from '/packages/dg-package-utils/main';
import { Utils } from '.meteor/local/isopacks/dg_utils/web.browser/main';
import { Utils } from '.meteor/local/isopacks/dg_utils/os/main';
```
When I want this:
```
import { Utils } from 'meteor/dg:utils';
```

trotzig:
I don't have any experience using meteor, and I haven't touched much of the meteor environment related code. Let's reach out to @rhettlivingston and see if he might be able to help out with your questions!

rhettlivingston:
Full disclosure, I haven't looked at this (or any) code in about 10 months and am having difficulty wrapping my head around it again.

In re: to [coagmano's last comment above](https://github.com/Galooshi/import-js/pull/436#issuecomment-309965574). It appears that the only place in meteorEnvironment.js that paths like '.meteor/local/isopacks/dg_utils/web.browser/main' exist is within the routine extractExportsFromMeteorPackage. That routine does not pass those paths out. I see no possibility of this code putting those paths into the namedExports array. The exports found are put into that array at [lib/environments/meteorEnvironment.js#L347](https://github.com/coagmano/import-js/blob/cff640b5d4678d875cd0ceaf2b2707702dd3a8c2/lib/environments/meteorEnvironment.js#L347). That line will only run once per package and will use the 'meteor/' form of the package name.

Therefore, I have to believe that the three choices being offered are coming from some other code.

Is it possible that the project's .meteor/ directory is being greedily scanned by other code that is automatically picking up the js files within it?

trotzig:
Thanks for replying @rhettlivingston! I'm sorry (?) we brought you back into the world of curly braces.  

> Is it possible that the project's .meteor/ directory is being greedily scanned by other code that is automatically picking up the js files within it?

Yes. Since a while back, all js files in the project are scanned for exports. 

@coagmano: could you try adding an `excludes` for that and see what happens? If it helps, we should probably ignore that folder by default.

coagmano:
Yeah I'll have another crack at it this week. 
It actually started working all of a sudden for me, so either way more testing (and tests!) are required

coagmano:
And one month later, I finally figured out the last test was failing because I wrote it badly haha.
I've been using this branch all month and It hasn't given me an isopack import path for awhile, so I think it's all good.

I haven't added tests for edge cases like badly behaved packages that don't include `fileOptions` in the `resources` array (ran into and fixed that one today, which triggered finishing this)

PR #437, opened 2017-06-21
Handle Flow type-alias exports and imports

rndmerle:
I hope I didn't forget anything, but it seems fine.
Indeed the case where one wants to export a class as a type, for instance, is not supported.
I mean, given an
`export default class TheClass { /* ... */ }`
you can't get
`import type TheClass from 'Module'`
because importjs would have to ask you each time how you want to import that class as.

Feel free to edit anything to your needs, or ask me to do so.

### DIFF
--- a/lib/Configuration.js
+++ b/lib/Configuration.js
@@ -39,7 +39,7 @@ const DEFAULT_CONFIG = {
   cacheLocation: ({ config }: Object): string => {
     const hash = crypto
       .createHash('md5')
-      .update(`${config.workingDirectory}-v3`)
+      .update(`${config.workingDirectory}-v4`)
       .digest('hex');
     return path.join(os.tmpdir(), `import-js-${hash}.db`);
   },
--- a/lib/ExportsStorage.js
+++ b/lib/ExportsStorage.js
@@ -9,9 +9,10 @@ function normalizedExportName(string) {
 }
 
 function normalizeRows(rows) {
-  // currently just casts `isDefault` to boolean
-  return rows.map(({ isDefault, ...other }) => ({
+  // currently just casts `isDefault` and `isType` to boolean
+  return rows.map(({ isDefault, isType, ...other }) => ({
     isDefault: !!isDefault,
+    isType: !!isType,
     ...other,
   }));
 }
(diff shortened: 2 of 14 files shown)


rndmerle:
I'll think about that today.
It happens I forgot to test again after the rebase and there are 2 easy fails I've to take care of.

rndmerle:
New commit to correct some falsy tests I forgot, and to add a few tests to Import-js : in the #fixImports section, mainly to test the interaction of mixed type/named/default imports. Not sure it's done right tho, and if it's enough.

I added a comment to `areOnlyTypes` but now I think of it, it's also possible to name that flag `isTypeStatement` or something along the idea that the statement itself is a 'type' one, leading to the type prefix in the `import type { }` statement. Is this better? Would it still need a little comment?

I added a comment to `areOnlyTypes` but now I think of it, it's also possible to name that flag `isTypeStatement` or something along the idea that the statement itself is a 'type' one, leading to the type prefix in the `import type { }` statement. Is this better? Would it still need a little comment?

About `isDefault` and `isType` you're right that an export can't be both afaik. So, because no enum available, it could be a `kind` integer? then normalized to `isDefault` and `isType` booleans.

lencioni:
For naming, we have been trying to use standardized language as much as possible. I notice when running imports through babylon that `import type foo` has `importKind: 'type'` and `import foo` has `importKind: 'value'`, so maybe something similar to that would be what we want here, similar to what you suggest.

trotzig:
> About isDefault and isType you're right that an export can't be both afaik. So, because no enum available, it could be an integer? named kind? then normalized to isDefault and isType booleans.

Yeah, that would work. I don't see this as high prio right now though. Let's keep it in mind if we do other changes. 

> I added a comment to areOnlyTypes but now I think of it, it's also possible to name that flag isTypeStatement or something along the idea that the statement itself is a 'type' one, leading to the type prefix in the import type { } statement. Is this better? Would it still need a little comment?

I think what you have now is fine.

trotzig:
Fantastic work @rndmerle, thanks for the contribution!

trotzig:
I'm going to await #436 for a while to see if things get resolved before cutting a new release.

Issue #415, opened 2017-06-07
findUndefinedIdentifiers returns a function that's referenced before it's declared.


dcporter:
I found an unsolved sub-case of #383: when the exported thing is a function. I dropped the following test into the `"knows about identifiers declared after usage"` block of [`findUndefinedIdentifiers-test.js`](https://github.com/Galooshi/import-js/blob/master/lib/__tests__/findUndefinedIdentifiers-test.js#L181), and it fails.

```js
  expect(
    findUndefinedIdentifiers(
      parse(
        `
    const abc = Item();
    function Item () {}
        `,
      ),
    ),
  ).toEqual(new Set([]));
```

(Edit: My test as originally posted exported `Item`; I re-ran without the `export` and it still fails.)

dcporter:
Followup: It's simpler than that, although the absolute reduced test case wanders into buggy-code territory:

```js
    var abc = count + 2;
    var count = 4;
```

This compiles (though of course it will result in `'undefined2'`).

Here's another example:

```js
    var abc = { def };
    function def() {}
```

dcporter:
@trotzig have you taken a look at [recast](https://github.com/benjamn/recast), or its kernel (basically) [ast-types](https://github.com/benjamn/ast-types)? It parses the ast like babylon (in fact using babylon I think?), but returns it wrapped in a `NodePath` thing, which makes it a lot easier to navigate, and comes with handy methods like [`hasBinding`](https://github.com/thejameskyle/babel-handbook/blob/master/translations/en/plugin-handbook.md#toc-checking-if-a-local-variable-is-bound), which I think might solve this whole thing out of the box. Doing this by hand is really prohibitively hard.

trotzig:
At the time I started looking into using a proper parser, I had very little experience with working with an ast. I looked at recast and a few other tools but I couldn't quite understand why I would need them. So I decided to do it by hand, to better understand. Had I started today, I would probably use a library though -- because you're right: it is hard.

Right now I don't have the bandwidth to take this on, but I'd love for someone to give it a try!

PR #443, opened 2017-07-03
Improve finding defined/undefined export variables

trotzig:


trotzig:
> You say you might want to refactor the current implementation soon. Any thoughts about the types of changes you want to make?
I think it would make sense to look at some library for the tree traversal. @dcporter outlines a few in #415. 

Hoisted imports are also covered. I'll add something to the test just to make sure we have it covered. Thanks for noticing!

PR #444, opened 2017-07-03
Improve messaging for added imports

trotzig:
Inspired by 3a84a27, I'm changing the message displayed when importing a
module. From

  Imported `./Foo` (main: index.js)

To

  Imported foo from './Foo'

I believe this makes the message easier to parse/understand for a js
developer.

### DIFF
--- a/lib/findJsModulesFor.js
+++ b/lib/findJsModulesFor.js
@@ -101,7 +101,7 @@ function dedupeAndSort(modules: Array<JsModule>): Array<JsModule> {
     sorted,
     (module: JsModule): string => module.importPath,
   );
-  return sortBy(uniques, (module: JsModule): string => module.displayName());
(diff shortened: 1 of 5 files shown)


lib/findJsModulesFor.js
```diff
@@ -101,7 +101,7 @@ function dedupeAndSort(modules: Array<JsModule>): Array<JsModule> {
     sorted,
     (module: JsModule): string => module.importPath,
   );
-  return sortBy(uniques, (module: JsModule): string => module.displayName());
+  return sortBy(uniques, (module: JsModule): string => module.importPath);
```
trotzig: At first, I was afraid this would change how we sort import statements, but after looking at how `displayName()` was implemented I believe this should be safe.

PR #449, opened 2017-07-19
Added example to configuration file section

deltaskelta:
I got confused on how to make the settings file with no example. Other projects using json settings files and changing states of javascript can lead to confusion about how to structure it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,8 +99,20 @@ the cursor on a variable and hit `<leader>g` (Vim), `(M-x) import-js-goto`
 
 ## Configuration
 
-ImportJS is configured through a JavaScript file (`.importjs.js`). Save the
-configuration file in the root folder of your project.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -99,8 +99,20 @@ the cursor on a variable and hit `<leader>g` (Vim), `(M-x) import-js-goto`
 
 ## Configuration
 
-ImportJS is configured through a JavaScript file (`.importjs.js`). Save the
-configuration file in the root folder of your project.
+ImportJS is configured through a JavaScript file (`.importjs.js`). 
+
+each setting should be a module export from the file like the example below:
```
trotzig: I'm not sure "module export" is the right term for this. How about this:

> The file needs to export a single object containing your configuration settings, like the example below:

?

Issue #445, opened 2017-07-06
Named and default export with the same name


rndmerle:
_Before considering working on a PR here is a potential improvement, but not sure if it's appropriate._

When using composition, a popular pattern is to name-export the base component and to default export the composed component, and this from a file with the same name (the purpose is to easily test the base component).
```
// Component.js
export const Component = {};
export default enhance(Component);
```

Then when you want to import the component, **most of the time you need the default import**:
```
import Component from './Component';
```

but it seems the **named import has priority** as we get:
```
import { Component } from './Component';
```

It could be considered as an anti-pattern I guess but it's quite popular nevertheless. I started to export that way when using HOCs.

So, what's the right behavior? Maybe asking which one should be imported?

coagmano:
I think there's trouble working out what the default export should be called. Officially it's name is `default` but you could keep track of the variable it was passed (if it's passed one) or use the file name?

trotzig:
Sorry for the late input here (I've been on a long vacation). The behavior I'm seeing is that both are presented to the user, so they can decide manually. Is that not the case for you?

rndmerle:
The idea would be to present both yea.
But again, it's just an idea because of the supposed popularity of this exporting pattern.

Currently, there is no choice, the named import is preferred, and there is no way both options to be presented to anyone because 1) one of the options is trimmed out by `dupeandsort()`, and 2) in the `unresolvedImports` array _(then  used for the choice popup on an editor add-on for instance)_, items are only differenciated by their `importPath`, there is no named/default specifier to differentiate the options.

trotzig:
Ah, I see what the problem is now. It's definitely a design issue with how the `unresolvedImports` interaction is with the editors. We need to send more information about the choice made by the user. 

This has come up before. In the beginning we used to just send the selected index from the `unresolvedImports` array. And then on the receiving end we would replay the entire import process, with modules already resolved (through their index). That was inefficient, and could potentially lead to mismatches (if things changed between the initial import and the replay). 

Because of this, I'd propose a solution that would make some information opaque to the client. Instead of passing down the `importPath`, we'd pass down an object that we received attached to the variable to resolve imports for. In json words, instead of this: 

```json
{
  "Button": [
    { 
      "displayName": "import { Button } from './Button'",
      "importPath": "./Button"
    },
    { 
      "displayName": "import Button from './Button'",
      "importPath": "./Button"
    },
  ]
}
```

We'd send something like: 

```json
{
  "Button": [
    { 
[...]

PR #451, opened 2017-08-01
Homonymous named/default export

rndmerle:
Ref: https://github.com/Galooshi/import-js/issues/445

To begin with, homonymous named and default imports are no longer trimmed down by `dedupeAndSort`, and the default import is placed before the named one in the list (up for debate).

Then, as you suggested, a `data` object is passed down with `unresolvedImports` so options can be differentiated.

`addImports()` is now accounting for this `data` object (still being backward compatible when receiving just a `importPath` string), and for instance uses `isNamedExport` property to return the right resolved import.

For further reference, just taking the case of the atom add-on, we'd have to modify this code:
```
askView.deferred
.then((resolved) => {
  const add = {};
  // add[word] = resolved.importPath;
  add[word] = resolved.data;
  addImports(add);
})
```

### DIFF
--- a/lib/__tests__/findJsModulesFor-test.js
+++ b/lib/__tests__/findJsModulesFor-test.js
@@ -2,8 +2,9 @@ import path from 'path';
 
 import Configuration from '../Configuration';
 import FileUtils from '../FileUtils';
+import JsModule from '../JsModule';
 import ModuleFinder from '../ModuleFinder';
-import findJsModulesFor from '../findJsModulesFor';
+import findJsModulesFor, { dedupeAndSort } from '../findJsModulesFor';
 
 jest.mock('../ModuleFinder');
 jest.mock('../FileUtils');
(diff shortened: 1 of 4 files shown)


lib/__tests__/findJsModulesFor-test.js
```diff
@@ -37,3 +38,20 @@ describe('when a matching module is excluded', () => {
...
+  ];
+
+  it('dedupes and sorts', () => {
+    expect(dedupeAndSort(modules)).toEqual([
+      new JsModule({ importPath: './bar', variableName: 'bar', hasNamedExports: false }),
+      new JsModule({ importPath: './foo/bar', variableName: 'bar', hasNamedExports: false }),
+      new JsModule({ importPath: './foo/bar', variableName: 'bar', hasNamedExports: true }),
+    ]);
+  });
+});
```
trotzig: Thanks for adding a test for this method!

rndmerle:
Indeed that's a shameful oversight. Thanks for spotting it.
Do you confirm that `filePath` is not needed as a root property? I don't know the ins and outs of various add-ons.

lib/__tests__/findJsModulesFor-test.js
rndmerle: It's a pleasure to test pure functions.

trotzig:
I've searched through all extensions and can't find any use of `filePath`.

PR #457, opened 2017-09-09
get extension from fs for relative local import

### LINKED ISSUE #455: Goto ignores extension for default export which does not match the filename
Given the file `./this-is-something-else.js` and code:
```javascript
import someWeirdName from './this-is-something-else';
```
It is navigated to `./this-is-something-else` (which does not exists) instead of navigating to file with `.js` extension. This might be related to #177

hontas:
fixes #455

### DIFF
--- a/lib/Importer.js
+++ b/lib/Importer.js
@@ -1,5 +1,6 @@
 // @flow
 import path from 'path';
+import fs from 'fs';
 
 import CommandLineEditor from './CommandLineEditor';
 import Configuration from './Configuration';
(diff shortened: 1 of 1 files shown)


trotzig:
Tusen tack! I'll give this a spin locally to see if I can come up with a test case. I have a different solution in mind that will simplify `goto` further, but this should at least help in the meantime.

trotzig:
I decided to merge this without a test. The one I could come up with was very tied to the implementation, so I'll punt on that until I have a more proper fix for `goto`.

PR #458, opened 2017-09-10
Make `goto` find more file paths

trotzig:
Before when running `goto`, we would ask the ModuleFinder for matches on
the variable name. We would then take that list and map it against
current imports. This only worked in cases where the file you were going
to was already in the index. When the file for some reason wasn't in the
index, goto wouldn't work.

To make things a little more robust, I'm changing the ordering of things
a bit. Instead of asking the index first, we look through current
imports and try to resolve the module name for the matching import
statement. If that fails (the variable isn't yet imported or the path
couldn't be resolved), we continue to ask the ModuleFinder for matches.

The new behavior is a little bit more robust, and slightly faster.

The changes made in #457 inspired me to dig a little deeper through
`goto` which led up to this commit.

### DIFF
--- a/lib/Importer.js
+++ b/lib/Importer.js
@@ -1,6 +1,7 @@
 // @flow
 import path from 'path';
-import fs from 'fs';
+
+import requireRelative from 'require-relative';
 
 import CommandLineEditor from './CommandLineEditor';
 import Configuration from './Configuration';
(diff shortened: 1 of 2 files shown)


trotzig:
Ping @hontas, this is basically your fix in a different form (using `require-relative` instead of looking directly in the file system).

trotzig:
The tests failed on Node v4 because it took too long. I retriggered and things were fine. I also made sure it passes locally on v4.

PR #462, opened 2017-09-30
Ignore packages labeled "isRoot": false

rivertam:
This was the approach I considered cleanest, but I'm definitely willing to implement something others might consider cleaner. I figured putting the root in the config would mean you'd have to make multiple configs, and that would create a lot of ambiguity.

### DIFF
--- a/lib/__tests__/findProjectRoot-test.js
+++ b/lib/__tests__/findProjectRoot-test.js
@@ -7,17 +7,24 @@ jest.mock('fs');
 
 afterEach(() => fs.__reset());
 
+const normalPackageJsonContents = `{
+  "name": "some-package",
+  "dependencies": {
+
+  }
+}`;
+
 it('finds the right folders', () => {
   // FileUtils.__setFile(path.join(process.cwd(), '.importjs.js'), {
-  fs.__setFile(path.join(path.resolve('/'), 'foo', 'package.json'));
-  fs.__setFile(path.join(path.resolve('/'), 'foo', 'bar', 'package.json'));
+  fs.__setFile(path.join(path.resolve('/'), 'foo', 'package.json'), normalPackageJsonContents);
+  fs.__setFile(path.join(path.resolve('/'), 'foo', 'bar', 'package.json'), normalPackageJsonContents);
   expect(findProjectRoot(path.join(path.resolve('/'), 'foo', 'bar', 'baz.js')))
     .toEqual(path.join(path.resolve('/'), 'foo', 'bar'));
 });
 
 it('treats package folders as roots', () => {
-  fs.__setFile(path.join(path.resolve('/'), 'foo', 'package.json'));
-  fs.__setFile(path.join(path.resolve('/'), 'foo', 'node_modules', 'bar', 'package.json'));
+  fs.__setFile(path.join(path.resolve('/'), 'foo', 'package.json'), normalPackageJsonContents);
+  fs.__setFile(path.join(path.resolve('/'), 'foo', 'node_modules', 'bar', 'package.json'), normalPackageJsonContents);
   // expect(findProjectRoot('/foo/node_modules/bar/baz/gaz.js')).toEqual(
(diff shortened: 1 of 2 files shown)


rivertam:
Yes, that sounds good. I figure lerna is a great usecase for this. Not sure how I'd go about merging the `package.json`s, though. To be honest, I don't know enough about import-js internals to understand entirely what it uses `package.json` for. Any guidance would be great.

Not sure what you mean by "inside a `importJS` object". Do you mean at runtime or do you mean in this library? Or do you mean a separate config file?

trotzig:
Yeah, the merging would be a little involved. Import-js reads `package.json` to figure out what's listed under `dependencies` and `devDependencies`. It then adds those dependencies to the index/cache so that you can quickly produce things like

```js
import React from 'react';
```

Merging these would mean refactoring the bit of code reading the package.json file. 

I meant that `isRoot` could be nested inside an `importJS`/`import-js` object, something like

```json
{
  "import-js": { 
    "isRoot": false
  }
}
```

rivertam:
re: `isRoot`, definitely a good idea. It might also be a good idea to support putting the entire config in there, but maybe that goes beyond the scope of this PR.

re: `package.json`, yeah that sounds good. It would definitely be pretty involved though because it would also require remembering which `node_modules` each dependency lies in, unless import already does something like it. (I'm at work so I don't have time to read through the code quite yet)

lencioni:
> It might also be a good idea to support putting the entire config in there

Possibly, but the config is JS, not JSON, so I'm not sure how useful this would end up being in practice.

rivertam:
True. Definitely one of my biggest gripes with npm. I know for my needs, I could definitely do it just using JSON.

rivertam:
Yeah that's smart. The only thing I'd be concerned about is picking up irrelevant `package.json`s and making the user think they have something installed if they don't and have an unconventional directory structure. But I think that's going to be the way to go; it's really low risk and generally easier to document. Thanks for the advice. I'll see if I can get to it tonight or by the end of the week.

trotzig:
> The only thing I'd be concerned about is picking up irrelevant package.jsons and making the user think they have something installed if they don't and have an unconventional directory structure

Agreed. Do you think stopping at something that has a `.git` folder is a good idea?

rivertam:
That sounds like a good compromise, although maybe we want to consider svn and mercurial as well, for those poor souls who can't use git

ivangeorgiew:
Is this functionality implemented? I really need that.

trotzig:
@ivangeorgiew There's been no more work in this area since this PR opened. If you want this right now, consider picking up where we've left off.

After reading up on this I realize there's another option as well. By overriding the `packageDependencies` function in your config, you can concatenate dependencies from multiple sources. The `meteor` environment utilizes this method, see https://github.com/Galooshi/import-js/blob/a6a7728f9d22c8b0e9fd485c8dfc058dc68c4e0a/lib/environments/meteorEnvironment.js#L418

ivangeorgiew:
Would this automatically work for finding exports from all the projects? Also I can't really get packageDependencies to work when I modify it, the way it is modified in meteorEnvironment.

trotzig:
> I have a project inside a project. The child project requires imports from the parent project. I tried renaming the workingDirectory in Configuration to the parent directory's destination as derived from process.cwd(), but I receive an error in ExportsStorage "Cannot read property 'all' of undefined" which refers to this.db, but here we enter the domain of sqlite3, which I sadly don't have any understaning about :D. Please Help!

@ivangeorgiew would you mind opening a separate issue for this? To help you out here, I would need more information about your setup. Folder structure would be helpful, and an example of how you want to import from one project to the other.

philolo1:
@ivangeorgiew @rivertam I am closing the issue, since this has been solved by my pull request https://github.com/Galooshi/import-js/pull/517 that was merged.

philolo1:
@trotzig since its not relevant anymore can you close it?

PR #472, opened 2017-11-30
remove extra , on tabJoined named exports

ivangeorgiew:


lib/ImportStatement.js
```diff
@@ -335,7 +335,7 @@ export default class ImportStatement {
     }
 
     const tabJoined = named.join(`,\n${tab}`);
-    const namedMultiLine = `{\n${tab}${tabJoined},\n}`;
+    const namedMultiLine = `{\n${tab}${tabJoined}\n}`;
```
trotzig: Most projects I work in prefer trailing commas. I often see this eslint rule turned on: https://eslint.org/docs/rules/comma-dangle

If we want to allow people to avoid dangling commas, we should add a new configuration option so that the default of adding a comma can be kept. Then, people can turn it off with config.

lib/ImportStatements.js
```diff
@@ -210,7 +210,7 @@ export default class ImportStatements {
             }));
         strings.push(...importStrings);
       });
-      strings.push(''); // Add a blank line between groups.
+      //strings.push(''); // Add a blank line between groups.
```
trotzig: If you don't want the spaces between groups, you can turn off grouping.

lib/ImportStatements.js
```diff
@@ -240,7 +240,7 @@ export default class ImportStatements {
       importsArray,
       (importStatement: ImportStatement): boolean =>
         !importStatement.isParsedAndUntouched(),
-    );
+    ).reverse();
```
trotzig: Nice! I saw you made a comment about this elsewhere. Although I would prefer for this to be either a separate commit in the PR or a separate PR altogether. So that we're not mixing concerns here.

PR #473, opened 2017-11-30
append imports option

ivangeorgiew:


README.md
```diff
@@ -318,6 +319,17 @@ disabled, existing imports are not rearranged, and new imports are always added
 sortImports: false
 ```
 
+### `appendImports`
+
+By default, ImportJS will add new imports below the existing imports.
```
trotzig: This is only true when `sortImports` and `groupImports` are false. I think the appending is fine to do without configuration so you can remove these docs and just change directly in the code. Sorry if I'm complicating things here.

LinuCC:
@trotzig Does this pull request still makes sense to you?
I'd like this feature, so I would like to continue work on it if possible.

ivangeorgiew:
@LinuCC this is so minor that I am going to include in my other PR - https://github.com/Galooshi/import-js/pull/506

PR #474, opened 2017-11-30
dangling commas option

ivangeorgiew:


README.md
```diff
@@ -329,6 +330,16 @@ importing. By default, only modules listed under `dependencies` and
 importDevDependencies: true
 ```
 
+### `dangleComma`
+
+By default, ImportJS will add trailing commas in named exports.
```
trotzig: Could you change this to `By default, ImportJS will add trailing commas when constructing import statements with multiple named imports. `?

README.md
```diff
@@ -329,6 +330,16 @@ importing. By default, only modules listed under `dependencies` and
 importDevDependencies: true
 ```
 
+### `dangleComma`
```
trotzig: I think it would be more natural for the config option to be named `danglingCommas`. It would make it more consistent with the rest of the config options.

lib/ImportStatement.js
```diff
@@ -37,6 +38,7 @@ export default class ImportStatement {
   originalImportString: ?string;
   leadingContent: ?string;
   path: string;
+  config: Configuration;
```
trotzig: Instead of passing in the whole config object, could you pass in only the new option? That would make the code a bit more explicit, and it's more in line with what we're doing elsewhere.

ivangeorgiew:
I don't know how to write the test and where in the file exactly. Maybe you can write it instead?

lib/JsModule.js
```diff
@@ -173,6 +173,7 @@ export default class JsModule {
       importFunction: config.get('importFunction', { pathToImportedModule }),
       namedImports,
       areOnlyTypes: this.isType,
+      danglingCommas: config.get('danglingCommas')
```
trotzig: This is what's causing the build failures. You need a comma at the end of the line.

ivangeorgiew:
Updated. You can pull it now.

trotzig:
Alright, let me look at a test after I merge this.

lencioni:
I think that trailing might be more consistent with other tools than dangling

trotzig:
Oh, I was going off of the eslint `comma-dangle` rule: https://eslint.org/docs/rules/comma-dangle

PR #475, opened 2017-11-30
empty line between groups option

ivangeorgiew:


ivangeorgiew:
Yes, In my project from work we group our imports, but don't sort them.
We don't add new line between groups however and don't sort them.
Imports are similar to variable declarations - you don't put new line between them, there is no point.
I think this option won't hurt even if it helps only 2-3 people.

trotzig:
Okay, fair enough. Thanks for explaining your use-case. The new config option in isolation doesn't add much overhead, but there's definitely a risk that this project ends up in a place where we have too many config options - which in turn can make the tool harder to use. That being said, I've seen this come up before and I'm okay adding this option. 

We do want some tests though, so I'll either follow up myself with a commit for that when I have time. Or if you figure out how to do it, I would be glad to review another PR for it. There might be some help from https://github.com/Galooshi/import-js/blob/master/CONTRIBUTING.md.

PR #477, opened 2017-12-01
Dont rearrange modified imports

ivangeorgiew:
When rearranging imports, because a new namedImport was added, merge conflicts can become confusing and

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ The following configuration options are supported.
 - [`stripFileExtensions`](#stripfileextensions)
 - [`tab`](#tab)
 - [`useRelativePaths`](#userelativepaths)
-- [`mergableOptions`](#mergableOptions)
+- [`mergableOptions`](#mergableoptions)
 
 ### `excludes`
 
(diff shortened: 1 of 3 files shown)


trotzig:
I understand the desire for this, and I’ve seen it come up in the past. But I’m not fully sold on adding this feature. I agree that merge conflicts can be confusing, but making import-js do less isn’t solving that issue. And even if this makes the number of merge conflicts go down ever so slightly, it’s a temporary thing. Once your organization has started adopting import-js, you will get fewer and fewer churn over time.

ivangeorgiew:
I can't force the people I'm working with to use import-js. Are there any benefits to moving the changed imports position? Imagine that you add one named import to 10 lines. It gets extremely confusing in the merge conflicts even someone else changed something too.

trotzig:
> I can't force the people I'm working with to use import-js.

I can totally understand that.

> Are there any benefits to moving the changed imports position? Imagine that you add one named import to 10 lines. It gets extremely confusing in the merge conflicts even someone else changed something too.

The benefit here is consistency. And while consistency can be a hurdle in adopting a new tool, I think you'll see long-term that being consistent about imports is better.

That being said, code churn can be an annoying issue. If you work on this PR (adding and fixing tests) I'm definitely willing to merge this in.

PR #486, opened 2018-03-16
Recognize destructured assignment to a new name (fixes #485)

### LINKED ISSUE #485: Object destruction with assigning to new names
If my code contains an object destruction with assigning to new constant name:

    const { id: langForeignID } = findByName(LANG_FOREIGN, en.langs);
    console.log(langForeignID);

Then import-js does not recognize this syntax and tries to import the already defined `langForeignID` constant:

    import { langForeignID } from "P/src/comp/learn/sel-domains";

This destruction syntax is a part of ES2015:
https://www.ecma-international.org/ecma-262/6.0/#sec-destructuring-assignment
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

kahless:
I also updated a `package.json` version: 2.10.0 -> 2.10.1
Tried this changes on my project with a full success. Hope this is going to work fine.

### DIFF
--- a/lib/__tests__/findUndefinedIdentifiers-test.js
+++ b/lib/__tests__/findUndefinedIdentifiers-test.js
@@ -25,6 +25,14 @@ it('recognizes destructured assignment', () => {
   `))).toEqual(new Set([]));
 });
 
+it('recognizes destructured assignment to a new name', () => {
+  expect(findUndefinedIdentifiers(parse(`
+    import Foo from 'foo';
+    const { bar: scar } = Foo;
+    scar();
+  `))).toEqual(new Set([]));
+});
+
 it('knows about scope', () => {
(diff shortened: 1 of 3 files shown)


kahless:
> I also updated a package.json version: 2.10.0 -> 2.10.1
But now I suppose it was a mistake. You can have a lot of pull requests and versions will clash. Sorry for that. It is one of my first pull requests.

trotzig:
> But now I suppose it was a mistake. You can have a lot of pull requests and versions will clash. Sorry for that. It is one of my first pull requests.

Don't worry about it, I'll handle that before publishing a new version. In genereal, it's best to leave version bumping to the maintainer/owner. Since, like you said, it's hard to keep track of all changes included in the next release for an outsider.

PR #487, opened 2018-03-16
Removed "namedExports" from deprecated options (fixes #484)

### LINKED ISSUE #484: lodash/fp problem (hacky exports)
There are packages that export names in a hacky way. For example "lodash/fp" exports this:

    var _ = require('./lodash.min').runInContext();
    module.exports = require('./fp/_baseConvert')(_, _);

import-js does not see this export and does not generate such imports:

    import { assoc, reduce, keys, compose, pull } from "lodash/fp";
    // import-js does not generate this!
    // No JS module to import for `...some lodash function name`

I can make a script like this to re-export functions and provide them to import-js:

    export { assoc, reduce, keys, compose, pull, ... } from "lodash/fp";

But such imports are not grouped correctly with other third party npm dependencies. "namedExports" property can be a workaround for this problem, but it is deprecated. Perhaps it should not be deprecated? Or am I missing something? I use lodash all the time so I need an automated tool for generating imports.

kahless:


lib/__tests__/Configuration-test.js
```diff
@@ -1019,13 +1019,14 @@ export { barbazCorge };
...
-        ]);
-      });
+      // Perhaps we will need this later.
+      // it('has a deprecation message', () => {
+      //   const configuration = new Configuration();
+      //   expect(configuration.messages).toEqual([
+      //     'Using namedExports to configure ImportJS is deprecated and will ' +
+      //     'go away in a future version.',
+      //   ]);
+      // });
```
trotzig: In general, it's better to just remove the code but I can see why you left this commented out here.

trotzig:
Ignore the build failure for now. I'll look into this if it continues to be an issue.

PR #492, opened 2018-04-02
Add util to namedExports for node

coaxial:
Relates to this issue: https://github.com/Galooshi/vim-import-js/issues/40

### DIFF
--- a/lib/environments/nodeEnvironment.js
+++ b/lib/environments/nodeEnvironment.js
@@ -43,4 +43,15 @@ export default {
     'vm',
     'zlib',
   ],
+  namedExports: {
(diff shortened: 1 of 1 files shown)


trotzig:
The build failures are on node 4, which has been failing for a while. They're not related to your change.

trotzig:
I'll hold off on publishing a new version for a while. Before that happens, folks can add this to their .importjs.js config to get the same behavior: 

```js
namedExports: {
  util: [
    'callbackify',
    'debugLog',
    'deprecate',
    'format',
    'inherits',
    'inspect',
    'promisify',
  ],
},
```

Issue #454, opened 2017-08-23
Compact groups


QuentinRoy:
I like the option to automatically group my imports by categories, but I am not a big fan of the whitespaces between groups. I haven't find a way to strip them and keep my imports grouped but compact. I wish we could have an option for this.

trotzig:
I agree that the current behavior is a little inflexible. Can you explain what you want the imports to look like? We've discussed making the groups configurable in the past and I think it's mostly waiting for someone to write a PR. I haven't seen whitespace removal come up before, but I'm sure you have good reasons.

QuentinRoy:
Simple, instead of this:

```js
import React, { Component } from 'react';
import propTypes from 'prop-types';

import { ContextSwitch, Context, contextPropType } from 'reacolo';

import ControlApp from './control/ControlApp';
import LoadingView from './LoadingView';
import XpApp from './display/XPApp';

class BaseApp extends Component {
  constructor(props) {
    super(props);
    /* stuff */
  }

  render() {
    return <div>stuff</div>;
  }
 }
```

I prefer this

```js
import React, { Component } from 'react';
import propTypes from 'prop-types';
import { ContextSwitch, Context, contextPropType } from 'reacolo';
import ControlApp from './control/ControlApp';
import LoadingView from './LoadingView';
import XpApp from './display/XPApp';

class BaseApp extends Component {
  constructor(props) {
    super(props);
    /* stuff */
  }

  render() {
    return <div>stuff</div>;
  }
 }
```

It might look less readable at with little below, but I don't find myself reading the import soon and while I like sorting the imports per categories, I don't feel it benefits much to add new lines between this categories. However I find it makes it harder to find the separation between the imports and the rest of the code (which for me is important).
Someone might also want to define what to separate or not, e.g. I could do with local import / global import, but the separation between named import and default import does not make much sense to me (in particular as they can be mixed, c.f. the first line).
My suggestion would actually to treat "grouping" and "alphabetical sort" as two different ways of sorting the imports: I find having two properties for this a bit confusing, what if I put "grouping" and "sorting" both to true ? Then deal with new line insertions as a separate option.

trotzig:
Thanks for providing that context. I agree that it's a little odd to have both those configuration options. I think the reason is that they were added one by one gradually, and we never took a holistic look at things. 

Providing a configuration option to disable the whitespace shouldn't be too hard I think. If that's something you'd find useful I'll be happy to review a PR.

Issue #448, opened 2017-07-12
append to import list?


cj:
When you have `groupImports: false` and `sortImport: false`, any new import is `prepended` to the top of the file.  If this was being done manually you'd normally always append it to your imports like:

```javascript
import old from 'old'
import new from 'new'
```

and not this, which is the current functionality:

```javascript
import new from 'new'
import old from 'old'
```

Maybe a new option called `newImportsDirection`, so you can choose?  i.e. `newImportsDirection: 'append'` would be a good way to solve this issue?

dabbott:
Maybe new imports should always be added to the bottom when sorting and grouping are turned off? I can't really see a use case for adding them to the top. I probably should've done this when I added the 'sortImports' option :P

ivangeorgiew:
In lib/ImportStatements.js reverse the result in the _toGroups() with .reverse()
   
   let result = partition(
      importsArray,
      (importStatement: ImportStatement): boolean =>
        !importStatement.isParsedAndUntouched(),
    ).reverse();

trotzig:
@ivangeorgiew do you want to push that as a PR? I'm happy to review and merge.

PR #506, opened 2018-07-03
Useful changes from my setup which I think may be beneficial to others

ivangeorgiew:


lib/ImportStatement.js
```diff
@@ -149,10 +149,7 @@ export default class ImportStatement {
    * @return {Array} an array that can be used in `sort` and `uniq`
    */
   toNormalized(): Array<string> {
-    if (!this.defaultImport && !this.hasNamedImports() && this.hasSideEffects) {
-      return [this.path];
-    }
-    return [this.defaultImport || '', ...this.localNames()];
+    return [this.defaultImport !== 'React', this.path];
```
trotzig: Why do you want `React` to be on top?

lib/visitIdentifierNodes.js
```diff
@@ -53,7 +53,7 @@ function normalizeNode(node, context) {
   }
 
   const isAssignment = KEYS_USED_FOR_ASSIGNMENT.has(key) ||
-    (key === 'key' && parent.parent.type === 'ObjectPattern') ||
+    (key === 'key' && !parent.computed && parent.parent.type === 'ObjectPattern') ||
```
trotzig: Can you come up with a test case demonstrating what this is supposed to fix?

lib/Watcher.js
```diff
@@ -66,34 +67,47 @@ export default class Watcher {
...
-            removed.push({ path: normalizedPath });
-          }
-        });
-        if (added.length) {
-          this.onFilesAdded(added);
-        }
-        if (removed.length) {
-          this.onFilesRemoved(removed);
-        }
+        findAllFiles(this.workingDirectory, this.excludes)
```
trotzig: While this may fix an issue for you, it's also going to slow down the file watcher significantly.

lib/Watcher.js
```diff
@@ -66,34 +67,47 @@ export default class Watcher {
...
+          .then((files: Array<Object>) => {
+            const added = [];
+            const removed = [];
+
+            files.forEach(({ path, mtime }: Object) => {
+              // add files which contain the following:
+              // export * from 'file'
+              if (fs.readFileSync(path, 'utf8').match(/export\s+\*\s+from/) !== null) {
```
trotzig: This should go in `findExports.js` instead of here.

lib/findProjectRoot.js
```diff
@@ -11,7 +12,12 @@ function findRecursive(directory) {
   const pathToPackageJson = path.join(directory, 'package.json');
 
   if (fs.existsSync(pathToPackageJson)) {
-    return directory;
+    const packageDescription = FileUtils.readJsonFile(pathToPackageJson);
+    const { isRoot = true } = packageDescription['import-js'] || {};
+
+    if (isRoot) {
+      return directory;
+    }
```
trotzig: This looks useful, and I'm assuming it's a continuation of https://github.com/Galooshi/import-js/pull/462? An update to `findProjectRoot-test.js` as well as an entry in the README and I think this is good to go. 👍

lib/Watcher.js
ivangeorgiew: This is an issue with everyone using `export * from './somePlace'`
I don't think there is any other way to do this.
Also please try it out - I don't notice any slowdown whatsoever.

trotzig: This method is called every time a change is made to a file (either modified, deleted or added). `findAllFiles` will traverse the filesystem, which will be slow on any large project. I think the reason you're not noticing this is that it's being done in the background. 

The `findExports` module already has code to handle `export ... from` declarations. Can you see if you can inject your logic there instead of here?
https://github.com/Galooshi/import-js/blob/a6a7728f9d22c8b0e9fd485c8dfc058dc68c4e0a/lib/findExports.js#L14

lib/visitIdentifierNodes.js
ivangeorgiew: added tests

ivangeorgiew:
I added tests for `isRoot` and info in the README

lib/Watcher.js
ivangeorgiew: Say you have a files `a.js`, `b.js`, `c.js`.
The contents of `a.js` are `const abc = someNewVar`
The contents of `b.js` are `export * from './c'`
The contents of `c.js` are `export someOldVar = 123`
When you add `export someNewVar = 321` in file `c.js`
and you try to add the import in `a.js` the only option will be `c.js`
since file `b.js` didn't change. This is an incorrect behavior.
We need to add `b.js` to the `processQueue` method in `ModuleFinder.js`
Since I don't know how to do that - I implemented the only solution I saw

lib/__tests__/findUndefinedIdentifiers-test.js
```diff
@@ -58,6 +58,13 @@ it('knows about dynamic keys', () => {
   `))).toEqual(new Set(['bar']));
 });
 
+it('knows about aliases in destructured objects', () => {
+  expect(findUndefinedIdentifiers(parse(`
+  const foo = { 'theValueOfMyConst': 123 };
+  const { [MY_CONST]: someAlias } = foo;
+  `))).toEqual(new Set(['MY_CONST']));
+});
```
trotzig: Nice.

lib/Watcher.js
trotzig: Okay, thanks for explaining the use-case. Still, we can't afford traversing through all files on every file modification. I think the way we should solve this is to introduce some kind of dependency list, so that we know to update exports for `b.js` when `c.js` is modified.

ivangeorgiew:
@trotzig I updated the PR, I don't think anything more has to be done. Can you please merge it now?

ivangeorgiew:
@trotzig What is the point of watching JSON files for changes btw? You can't import from json files as far as I know.

lib/Watcher.js
```diff
@@ -68,6 +68,21 @@ export default class Watcher {
...
+        //       // add files which contain the following:
+        //       // export * from 'file'
+        //       if (!path.includes('node_modules')
+        //         && fs.readFileSync(path, 'utf8').match(/export\s+\*\s+from/) !== null
+        //       ) {
+        //         added.push({ path, mtime, exportsAll: true });
+        //       }
+        //     });
+        //   });
+
```
trotzig: This commented-out code isn't adding anything at the moment. I suggest opening a separate issue explaining what change in behavior you'd like to see.

lib/ImportStatement.js
trotzig: I'm not sure I understand. I haven't come across any project where React has to be at the top.

lib/ExportsStorage.js
```diff
@@ -136,8 +135,8 @@ export default class ExportsStorage {
           items.forEach(({ path: pathToFile, mtime }) => {
             mtimes[pathToFile] = mtime;
           });
-          const filtered = files.filter(({ path: pathToFile, mtime }) => (
-            mtime !== mtimes[pathToFile]
+          const filtered = files.filter(({ path: pathToFile, mtime, exportsAll }) => (
+            mtime !== mtimes[pathToFile] || exportsAll
```
trotzig: Revert the changes made to these two lines. The `exportsAll` option is not in use after you cleaned out that logic.

ivangeorgiew: done

lib/Watcher.js
ivangeorgiew: done

README.md
```diff
@@ -117,6 +117,18 @@ Save this file in the root folder of your project (e.g. where the package.json
 file is found). You can also save it to the user home directory if you want to
 share a global config between different projects.
 
+If you have a project inside a project with both containing a `package.json` file,
+you can specify that the subproject is not root like this:
```
lencioni: Because most of the configuration for import JS is in the import JS configuration file, I think it would be helpful to make it clearer that this snippet is showing `package.json`. Maybe all we need to do is add "...not root like this in `package.json`:" or something

lib/ImportStatement.js
```diff
@@ -148,11 +148,8 @@ export default class ImportStatement {
   /**
    * @return {Array} an array that can be used in `sort` and `uniq`
    */
-  toNormalized(): Array<string> {
-    if (!this.defaultImport && !this.hasNamedImports() && this.hasSideEffects) {
-      return [this.path];
-    }
-    return [this.defaultImport || '', ...this.localNames()];
+  toNormalized(): Array<any> {
+    return [this.defaultImport !== 'React', this.path];
```
lencioni: Can you help me understand why this change is made (including the type change)? It seems weird to have React hard-coded in this project.

lib/ImportStatement.js
ivangeorgiew: I guess I was wrong - it is only standard in the firm I work in. Therefore I left it to [this.path] only.

lib/ImportStatement.js
ivangeorgiew: I reverted the change. Now it is only [this.path]

README.md
ivangeorgiew: done

ivangeorgiew:
I believe I made all the changes
Can you merge now?

ivangeorgiew:
added logic for initializing the module finding when using an "init" function>
the necessary change was made in vim-import-js as well: https://github.com/Galooshi/vim-import-js/pull/43

lib/ImportStatement.js
```diff
@@ -149,10 +149,7 @@ export default class ImportStatement {
    * @return {Array} an array that can be used in `sort` and `uniq`
    */
   toNormalized(): Array<string> {
-    if (!this.defaultImport && !this.hasNamedImports() && this.hasSideEffects) {
-      return [this.path];
-    }
-    return [this.defaultImport || '', ...this.localNames()];
+    return [this.path];
```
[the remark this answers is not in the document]
trotzig: There has been plenty of discussion around sorting and grouping, and it's pretty clear that we need more flexibility: 
https://github.com/Galooshi/import-js/issues/508
https://github.com/Galooshi/import-js/issues/454
https://github.com/Galooshi/import-js/issues/448
https://github.com/Galooshi/import-js/issues/284

I think the best thing we could do is to look at what the sort-imports eslint plugin does and make things configurable the same way. 

I'm not going to argue much with what you think is the right order. What's important here is that we can either introduce a breaking change (that doesn't fix the problem) or keep the current sorting until we have a non-breaking and flexible solution. Introducing a breaking change on import sort order right now would mean introducing churn for every project that has been using import-js for a while.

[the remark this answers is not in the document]
ivangeorgiew: Two of those issues are things that I fixed myself - the empty line between imports being removed and the newly added imports being appended instead of prepended.
The other two issues are from people which want exactly the same sort order as I have currently implemented in this PR.
So I don't really see any conflicts of interest... Everyone wants the same thing.
Those issue links just highlight what I have told you in my previous comment.

lib/ImportStatements.js
```diff
@@ -243,7 +243,8 @@ export default class ImportStatements {
       importsArray,
       (importStatement: ImportStatement): boolean =>
         !importStatement.isParsedAndUntouched(),
-    );
+    ).reverse();
```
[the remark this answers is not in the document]
ivangeorgiew: https://github.com/Galooshi/import-js/issues/448
This is one of the issues you have commented in the file above.
Look at the bottom.

lib/ImportStatement.js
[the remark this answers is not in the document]
trotzig: > I really don't see any other possible sorting as viable

Let me present you with a few possible sorting options that are all equally viable: 

- a-z on the variable name (the assignment)
- a-z on the module name (the path/package name)

For groups, any combination of the following seems equally viable: 
- built-in packages
- node_modules
- non-js files
- side-effect imports (those that have no assignment)
- relative paths
- absolute paths

The right thing here would be to make things configurable, so that you can choose any sorting and any grouping. If we make the default config produce the same sorting and grouping result as what we have today by default, I'll be happy to merge.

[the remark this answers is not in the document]
ivangeorgiew: First option:
> a-z on the variable name (the assignment)

This is not viable, because of I what I wrote in the first comment...
Since you haven't read it I will quote it here...

> Sorting by the imports themselves and not where they come from doesn't make any sense. This will produce many unnecessary/confusing merge conflicts. Example: instead of only removing a named export, the change will also include the movement of the imports for that file due to the sorting. The same scenario will be for appending named imports.

Second option:
> a-z on the module name (the path/package name)

This is what I have currently implemented...

For the grouping, it is very simple - exports from outside of the project, after which - local exports. I think everyone will find that as a reasonable default.

If you find any value in adding a custom sorting/grouping function that will benefit anyone (since I haven't seen one such person in the above mentioned issues by you),
then do that yourself. I have done enough work, I am happy with what I've got and I consider it to be the best default behavior for the project.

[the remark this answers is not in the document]
coagmano: I currently have a few projects using the current default sorting and grouping rules and I agree that this shouldn't be changed until it's configurable. Otherwise I'll need to turn it off and manually sort and group again, or deal with PR churn for the foreseeable future. I'm sure there are others using the default behaviour that we don't hear from.

Because this is an area where it's down to preference, and an existing default has been in use for a fairly long time, it does not make sense to change the default and push a different preference on those already using the default behaviour.

Now @trotzig has _not_ asked you to implement custom sorting configuration, only to keep the existing defaults so it doesn't create churn for other users.
I understand the frustration of having a PR held up when you feel that you have done a good job, but lets keep some perspective.

Also, you have done a great job here. As a user I always appreciate when someone makes an effort to improve OSS tools for the community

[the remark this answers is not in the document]
ivangeorgiew: The current setup doesn't make sense for the reason I have pointed out above. That is why I changed it to something that does. You simply don't want the sorting to be changed because that is what you have used in existing projects. I consider this argument invalid. Keeping wrong code because it is what people have used so far is not correct. Please give me an argument that supports the claim that the current sorting is in any way better than the one I have implemented.
If you can't, someone else will have to implement these features listed here, because I got tired of this.

PR #505, opened 2018-06-22
Babel7

paradoxxxzero:
This Pull Request is not aimed to be merged as-is.

Simply, I ported this project to babel 7 beta to enjoy new syntaxes (noticeably the new react Fragment one which is tested now), and I thought you could be interested.

Thanks for this very useful project.

PR #509, opened 2018-07-29
Add TypeScript support, bump to babel v7

williamboman:
I've included stuff from #505.


williamboman:
@lencioni This should be ready for 👀 now

lib/parse.js
```diff
@@ -1,18 +1,42 @@
-export default function parse(fileContent: String): Object {
-  // For some reason, we can't `import` babylon using an es6 import.
+const TYPESCRIPT_FILE_PATH_REGEX = /\.tsx?$/
+
+export default function parse(fileContent: String, absolutePathToFile: String): Object {
```
williamboman: Not the cleanest function signature (adding `absolutePathToFile: String`). I'm too lazy to come up with a better solution. Ideally I feel like this module should not incorporate any concept of files at all.

package.json
```diff
@@ -43,33 +45,36 @@
   },
   "homepage": "https://github.com/galooshi/import-js#readme",
   "devDependencies": {
-    "babel-cli": "^6.26.0",
-    "babel-core": "^6.26.0",
-    "babel-eslint": "^8.0.0",
-    "babel-jest": "^21.0.2",
+    "@babel/cli": "^7.0.0-beta.51",
```
lencioni: Do you mind bumping this to the latest beta? (currently beta 55 it seems)

lib/parse.js
lencioni: Agreed. I wonder if this would be better as an options object with a key of `typePlugins`?

lib/parse.js
```diff
@@ -1,18 +1,42 @@
...
+  // For some reason, we can't `import` @babel/parser using an es6 import.
   // eslint-disable-next-line global-require
-  const babylon = require('babylon');
+  const babelParser = require('@babel/parser');
+
+  const typePlugins = TYPESCRIPT_FILE_PATH_REGEX.test(absolutePathToFile)
+    ? ['typescript']
+    : ['flow']
```
lencioni: Not a big deal either way, but why use an array here instead of a string?

lib/parse.js
williamboman: That's exactly what I thought as well, and is what I did initially. However, I opted for the more rudimentary solution as I feel an options object parameter just introduces bloat and will make things harder to refactor. I'm up for either though 🤷‍♂️

lib/parse.js
williamboman: Just lazy coding 🙈, was unsure how many plugins was required for TS so I coded it as an array before figuring it out.

yarn-error.log
```diff
@@ -0,0 +1,131 @@
+Arguments:
```
lencioni: It looks like an error log snuck in. Mind removing this from your PR?

williamboman: Woops, done. Added some stuff to `.gitignore` while I was at it

package.json
```diff
@@ -43,33 +45,36 @@
   },
   "homepage": "https://github.com/galooshi/import-js#readme",
   "devDependencies": {
-    "babel-cli": "^6.26.0",
-    "babel-core": "^6.26.0",
-    "babel-eslint": "^8.0.0",
-    "babel-jest": "^21.0.2",
+    "@babel/cli": "^7.0.0-beta.55",
```
lencioni: It looks like beta 56 was published last week. Mind updating again?

williamboman: The semver range will include it and any later beta releases

trotzig: I still got beta.55 when I installed. Are you sure semver ranges work with pre-releases like these? Nothing to worry about though, I'm going to bump us to rc.1 anyway. :)

williamboman: It should, it should actually also include release candidates.

PR #514, opened 2018-08-15
Implement support for CoffeeScript 2

alin23:
Now that CoffeeScript 2 is released and stable, I think we should support it too.

There's no Babel plugin for transpiling CoffeeScript because it doesn't have a Babel compatible AST. But thankfully, CoffeeScript offers a `compile` function that gives us Babel-parseable JS code.


trotzig:
Nice! Not that I'm a coffescript user myself, but I'm sure this can help a lot of people out. Just to make sure I understand this right -- this PR only adds support for importing coffee files, it doesn't make it possible to use import-js in coffescript files. Right? You might want to clarify that somewhere  in the Readme.

alin23:
It actually adds complete support for CoffeeScript. With these changes I can call the CLI on `.coffee` files and I can use the Sublime integration inside `.coffee` files.

The only thing it doesn't is to remove trailing semicolons (which are illegal in CoffeeScript). This is something the users should configure through `importStatementFormatter`. I'll add a README section about this specific caveat and it should be good to go. What do you think?

trotzig:
There are some linting issues. I'll fix those up before I merge this.

trotzig:
I fixed the lints and merged this to the [`coffeescript-support`](https://github.com/Galooshi/import-js/tree/coffeescript-support) branch for now. A pre-release was also shipped that you can install by running `npm install -g import-js@next`. I tried running some tests locally and I was noticing some weird behavior when using the `fix` command - imports weren't being placed at the right place (they were always at the top). Will you run this version for a while and see if you can spot that (and other bugs) before we make a proper release?

alin23:
Thanks for making the pre-release!

I'm not using the sort functionality because I'm using a different tool to sort and group imports in CoffeeScript. That's why I didn't notice the problem you mentioned.

But I think this happens because imports in CS are not recognized by ImportJS when looking where to insert the new imports. I'm not sure how it could do it since we only have a way to transpile JS to CS, and not CS to JS.

I'm using this tool daily and it's already become a major time-saver for me, so I'll have enough occasions to spot the bugs. I'll report back here and bring fixes where I can. 

I'm still wrapping my head around the ImportJS codebase, it's bigger than I expected and not that easy to follow because of all the `Promise` constructs. I'm more used to the `async/await` syntax.

trotzig:
Yep, there’s some legacy here for sure. The tool was originally written in
Ruby, then ported to javascript.

A quick theory that might be worth exploring: when you compile to js from
coffee, is there a way to preserve line-breaks? I think that’s what’s
causing the sorting to be wrong. It’s all in one line after compilation.
Note that this is just a theory though.

alin23:
@trotzig I'll look into it, you may be right about the line breaks.

mikabytes:
I'm closing this as it has not seen any activity for a long time.

If it is still relevant, please post a comment and we'll reopen it. Thanks!

PR #527, opened 2019-01-17
Allow imports to be appended.

joeheyming:
Currently, if you have sortImports false, and groupImports false,
you will have imports appear at the top of the import list.

This is not ideal.
Now, there is a path forward which puts imports last instead of first.

### DIFF
--- a/lib/ImportStatements.js
+++ b/lib/ImportStatements.js
@@ -244,6 +244,11 @@ export default class ImportStatements {
       (importStatement: ImportStatement): boolean =>
         !importStatement.isParsedAndUntouched(),
     );
+
+    // appendImport only works when sortImports and groupImports is false
+    if (this.config.get('appendImports')) {
+      result.reverse();
+    }
     result = flattenDeep(result);
 
     if (this.config.get('sortImports')) {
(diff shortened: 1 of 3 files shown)


felixaa (approved):

joeheyming:
seems I don't have write access, so someone else will have to merge the PR

felixaa:
Hi! Can someone please have a look and merge this? I don't see any reasons why new imports should be added to the top

lib/ImportStatements.js
```diff
@@ -244,6 +244,11 @@ export default class ImportStatements {
       (importStatement: ImportStatement): boolean =>
         !importStatement.isParsedAndUntouched(),
     );
+
+    // appendImport only works when sortImports and groupImports is false
+    if (this.config.get('appendImports')) {
+      result.reverse();
```
lencioni, line 250: Does this have the effect of reversing all of the imports? Can you add a test that includes more than one existing import please, to show that the list is appended to and not just reversed?

mikabytes:
I'm closing this as it has not seen any activity for a long time.

If it is still relevant, please post a comment and we'll reopen it. Thanks!

Issue #546, opened 2019-12-04
Use ImportJS without package.json


mikabytes:
We use native ES modules and have no need for a bundler such as webpack. I would like to use this awesome tool anyway. However, it doesn't work because ImportJS have to have a parent directory with a package.json file.

The current workaround is to add the package.json just to fool the tool. Could we maybe add an option to instead look for a `.git` folder?

trotzig:
That sounds like it could work. Or we could allow people to drop a `.importjs-root` (or something) file in the directory structure. The `.git` approach is likely a breaking change, but the `.importjs-root` approach would be backwards-compatible. 

If you end up starting a PR in this area, here's a good place to start: https://github.com/Galooshi/import-js/blob/b9e202ad3fb6962273394d03b2153fd39030bb88/lib/findProjectRoot.js

mikabytes:
Thanks for pointing me in the right direction. I think I got good PR for it, but let me know if I missed some convention. I'll give it another go.

trotzig:
Closed by c93ce12

trotzig:
I've started working on releasing v4. There are a few other changes I want to make as part of that change but I'll release it as soon as it's ready.

PR #547, opened 2019-12-05
Add option to use .importjs-root file to detect project root

mikabytes:
As discussed in https://github.com/Galooshi/import-js/issues/546

This adds the option to create a file `.importjs-root` to allow for ImportJS to detect the root of the project.

This is useful for projects that don't have a `package.json`, such as those that use native ES modules only.

### DIFF
--- a/lib/__tests__/findProjectRoot-test.js
+++ b/lib/__tests__/findProjectRoot-test.js
@@ -14,6 +14,8 @@ const normalPackageJsonContents = `{
   }
 }`;
 
+const emptyFile = '';
+
 it('finds the right folders', () => {
   // FileUtils.__setFile(path.join(process.cwd(), '.importjs.js'), {
   fs.__setFile(path.join(path.resolve('/'), 'foo', 'package.json'), normalPackageJsonContents);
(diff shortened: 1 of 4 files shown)


mikabytes:
Hm, could someone help me understand what's going on with Travis? Those errors seem unrelated to my changes, and all tests pass when I run them locally.

trotzig:
Looks like failures are coming from Node v6&7. Are you able to repro locally with one of those versions? If not I think it’s safe to cut out Node < 8 (I can make that happen as part of releasing this change).

trotzig:
Btw, thanks for making a PR! Quick turnaround! 👏

mikabytes:
Happy to contribute to such an awesome tool. One of the few I use every hour of the day ;)

I have confirmed this is an issue with Node 7 (probably also holds true for 6). Also confirmed that the same test results in master branch, so this is not related to this PR. 

Always happy to drop support. You have my encouragement :)

trotzig:
After some CI/build issues I was finally able to get a release shipped. This change is included in v4. 🚢

trotzig:
What editor are you using? I'm a vim user myself, but I made sure to bump the Atom plugin as well (which is the only plugin with a direct dependency on the import-js npm lin--others rely on a globally installed command).

mikabytes:
Awesomeness! Quick turnaround indeed.

I'm a devout Vim user, and although I don't like to bloat my editor with YouCompleteMe and such, ImportJS is really bringing that warm breeze into my work environment. Especially when working with libs that require lots of small little imports; Redux Saga, Yup, etc.

_Ha en fin vinterdag! Snart läggdags här i Filippinerna ;)_

Issue #550, opened 2020-02-14
Add CustomElements support


mikabytes:
Perhaps this idea is a bit esoteric, but consider this situation:

### ./file1.js
```
class MyCustomElement extends HTMLElement {
 ...
}

window.customElements.define("my-custom-element", MyCustomElement)
```

### ./file2.js
```
const mydom = html`<my-custom-element>Hello</my-custom-element`
document.body.append(mydom)
```

I'd like ImportJS to add `import './file1.js'` to file2.js

The html`` template literal syntax is widely adopted in JavaScript community. Used by LitHtml, with support by Prettier, syntax support is available for Vim, Visual Studio and probably many others. If we could teach ImportJS to understand them as well, it would be a great addition and a huge win for me :)

trotzig:
I think this makes sense. I personally don't have a need for this but I'd be happy to help reviewing a PR! There are two parts to this. First, we need to make sure the "export" syntax in `./file1.js` is recognized and indexed by import-js ([here I think](https://github.com/Galooshi/import-js/blob/master/lib/findExports.js)). Then we need to find custom elements like the ones in `./file2.js`, likely here: https://github.com/Galooshi/import-js/blob/master/lib/findUsedIdentifiers.js

mikabytes:
Thanks for the pointers! I might start work on this already tomorrow if nothing else comes up.

I think we also need a configuration option to enable/disable this feature. Some people like to import their custom elements in a separate file (custom elements is a global registry)

mikabytes:
It's been 2 years.

While I did implement a solution that worked for my use case, it was evident in my company that it was too opinionated to be made into a general solution. People name their custom elements based on their convention. 

In the end, I had an idea for a more general plugin-style extension for ImportJS but never got around to writing it.

Closing this due to it being stale. Might reopen in the future.

PR #551, opened 2020-02-14
Add support for adding custom elements

mikabytes:
Based on #550, this PR indeed adds support for importing files that define custom elements. It has a few quirks though.

### Quirks 
- I want it to be treated as a named export, but JavaScript doesn't allow me to specify such metadata. The consequence is that imports can be added, but not matched or removed.
- Correct me if I'm wrong, but it seems ImportJS doesn't have support for "that which has a name but is not a named export". My solution is horrible, adding it as a named export and then stripping it away as a side-effect later.
- I lacked the energy to write tests for this. I will, but first, I would like to invite your comments on what I've done so far.
- I'm not too happy about the way I'm checking the type-of-export based on the adding <> characters to the name. I AM happy about using <> to have ImportJS give me distinct messages about importing a custom element though.
- I didn't find a suitable way to traverse the AST to find all the statements, so I got lazy and used a regular expression instead. This means that it will incorrectly import files where this statement has been commented out.
- No configuration option to disable this feature (yet)

### One idea
Maybe we could add support for "that which has a name but is not named in JS spec" as a comment? Maybe something like:

```
import './my-custom-element.js' // <my-custom-element>
```

OR

```
import /* <my-custom-element> from */ './my-custom-element.js'
```

Separating multiple named exports with comma. Or something. I'm happy to hear your ideas.

### other
Not too bad for 3 hours work on a foreign codebase.... :)

### DIFF
--- a/lib/findProjectRoot.js
+++ b/lib/findProjectRoot.js
@@ -12,7 +12,7 @@ function isRootPackageJson(path) {
 
   const packageJson = FileUtils.readJsonFile(path);
 
-  if ('importjs' in packageJson && 'isRoot' in packageJson.importjs) {
+  if (typeof packageJson === 'object' && 'importjs' in packageJson && 'isRoot' in packageJson.importjs) {
     return packageJson.importjs.isRoot;
   }
 


trotzig:
Apologies for the delay here! I hope to have a look at this later today.

lib/findExports.js
```diff
@@ -404,6 +404,11 @@ function findRootNodes(ast) {
   return realRootNodes;
 }
 
+function findUnnamedExports(data) {
+  const regex = /customElements\.define\s*\(\s*['"]([0-9a-z-]+)['"]\s*,\s*[a-zA-Z0-9_$]+\s*\)/g
+  return Array.from(data.matchAll(regex)).map(t => `<${t[1]}>`)
+}
+
```
trotzig: The regex here is fine, but make sure to add tests that show the behavior in different edge-cases so that they're clearly documented (like the commented-out code case you mentioned).

trotzig:
Excellent work!

> I want it to be treated as a named export, but JavaScript doesn't allow me to specify such metadata. The consequence is that imports can be added, but not matched or removed.

You're right that ImportJS doesn't have support for that, but you're wrong about your solution being horrible. :) 

> I'm not too happy about the way I'm checking the type-of-export based on the adding <> characters to the name. I AM happy about using <> to have ImportJS give me distinct messages about importing a custom element though.

I think this is fine. 


> `import /* <my-custom-element> from */ './my-custom-element.js'`

I like this idea. If you can make this work, go for it. If not, I think it's fine if removing the imports is a manual process.

lib/visitIdentifierNodes.js
```diff
@@ -102,10 +116,9 @@ export default function visitIdentifierNodes(
       queue.unshift(...itemsToAdd);
       continue; // eslint-disable-line no-continue
     }
-    const normalizedNode = normalizeNode(current.node, current.context);
-    if (normalizedNode) {
+    normalizeNode(current.node, current.context, (normalizedNode) => {
       visitor(normalizedNode, current.context);
```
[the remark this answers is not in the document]
mikabytes: > optionally returning an array
Having mixed return values have a bit of a code smell for me, having to test the object in order to use it. This way the function is more honest: "I'm going to produce zero, one or several normalizedNodes".

Of course, we could always return an array. Not sure why I went with the callback pattern. I don't like it, I never liked it :) An iterable of some kind makes the most sense. I'll convert it to a generator if you agree this would be a most pleasant way forward?

lib/findExports.js
mikabytes: I gave this some more thought and realized the problems don't end at commented code. It would also pick up the pattern inside of strings etc.

I'm becoming more certain that traversing the AST tree is the proper way to go. Do you have any indication on how much work that would involve? Guess it couldn't be too difficult?

mikabytes:
> but you're wrong about your solution being horrible

Haha, agree to disagree ;-)

> import /* <my-custom-element> from */ './my-custom-element.js'

I'll have something ready once I have some time to do it. Might not be this week, we'll see!

lib/visitIdentifierNodes.js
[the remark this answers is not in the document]
trotzig: I don't have a strong preference, you make the call :)

lib/findExports.js
trotzig: Hard to say, these things tend to be very explorative in nature. I usually TDD things like this, start with adding a test case to `findExports-test.js` and iterate from there. 

The AST traversing helpers in this library aren't super intuitive, and there might be better alternatives out there. But with a TDD approach things tend to work quite well.

PR #554, opened 2020-03-19
add example for namedExports

### LINKED ISSUE #553: namedExports solved the 'No JS module to import for Redirect'  problem
> Since 2.1.0 ImportJS finds your named exports automatically. Most likely you don't need this option.

Apparently I need this for some modules. For example { connect } from 'react-redux', React from 'react' libraries and others imported as expected. But things from 'react-router-dom' gave me a headache =).
My package.json is valid and has 'react-router-dom' as a dependency library. The solution is
```js
module.exports = {
  excludes: ['./public/**'],
  emptyLineBetweenGroups: false,
  sortImports: false,
  groupImports: false,
  minimumVersion: '4.0.2',
  namedExports: {
    'react-router-dom': ['Route', 'Redirect']
  }
};
```
I don't know what is special with the particular library but with the settings importing works.

dagolinuxoid:
closes #553

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,6 +266,40 @@ The key used to describe the named exports should be a valid import path. This
 can be e.g. the name of a package found under `node_modules`, a path to a
 module you created yourself, or a relative import path.
 
+Consider the example as a valid use case for the `namedExports` property. Let's say we have a file:
+```jsx
+import { Provider } from 'react-redux';
+import React from 'react';
+import store from './redux/redux-store';
+import ReactDOM from 'react-dom';
+import App from './App';
+
+ReactDOM.render(
+  <BrowserRouter>
+    <Provider store={store}>
+      <App />
+    </Provider>
+  </BrowserRouter>,
+  document.getElementById('root')
+);
+
+```
+And we're going to import `BrowserRouter` but instead of the desired result we get
+the *No JS module to import for `BroswerRouter`* message.
+In order to fix the problem populate `namedExports` of your config file as follows:
+```js
+module.exports = {
+  excludes: ['./public/**'],
+  namedExports: {
+    'react-router-dom': ['BrowserRouter', 'Route', 'Redirect']
+  }
+};
+
+```
+After that we get the line attached to our file on importing as expected:
+
+`import { BrowserRouter } from 'react-router-dom'`
+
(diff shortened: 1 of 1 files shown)


dagolinuxoid:
I didn't alter existing text of docs but rather appended my own. I believe it's not that bad  =)

README.md
```diff
@@ -266,6 +266,40 @@ The key used to describe the named exports should be a valid import path. This
...
+    <Provider store={store}>
+      <App />
+    </Provider>
+  </BrowserRouter>,
+  document.getElementById('root')
+);
+
+```
+And we're going to import `BrowserRouter` but instead of the desired result we get
+the *No JS module to import for `BroswerRouter`* message.
```
trotzig: small typo here on `BrowserRouter`

README.md
```diff
@@ -266,6 +266,40 @@ The key used to describe the named exports should be a valid import path. This
...
+In order to fix the problem populate `namedExports` of your config file as follows:
```
trotzig: A comma after "problem" will make this sentence a little easier to read. Also, change "of your" to "in your".

README.md
```diff
@@ -266,6 +266,40 @@ The key used to describe the named exports should be a valid import path. This
...
+```js
+module.exports = {
+  excludes: ['./public/**'],
```
trotzig: This line could be deleted.

README.md
```diff
@@ -266,6 +266,40 @@ The key used to describe the named exports should be a valid import path. This
...
+  namedExports: {
+    'react-router-dom': ['BrowserRouter', 'Route', 'Redirect']
+  }
+};
+
+```
+After that we get the line attached to our file on importing as expected:
```
trotzig: How about "After that we are able to import `BrowserRouter` correctly. The resulting import statement will look like this:"?

README.md
dagolinuxoid: yep, I noticed  that happens especially often when you're making changes in a browser rather than in your editor locally

README.md
dagolinuxoid: sure

README.md
dagolinuxoid: sounds good

PR #580, opened 2021-12-03
Does not import React into JSX when version is gte 17

madox2:
As of react 17.0.0 it is not needed to import React into JSX files: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html. Added functionality to conditionally import React based on the react version in the package.json.


mikabytes:
I'm closing this as it hasn't seen any activity for a long time. But I started an issue #597 to discuss it, and if needed, we'll reopen this at a later date.

PR #588, opened 2024-01-11
Upgrade sqlite3 to v5

mikabytes:
The old v4 sqlite3 requires node-gyp to build from source, which doesn't compile on a modern machine. I'm not sure as to exactly why, just that the general recommendation is to upgrade to v5 which uses Node API bindings.

This PR does exactly that. I haven't tested this extensively, just used it as usual in my projects, and it seems to be fine.

Additional testing would be appreciated.

It's been a few years since import-js made a release. Since then the package lockfile format has changed and is automatically updated. This resulted in a huge changeset for this PR. I promise I haven't changed anything else than  sqlite3 though :-)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,16 +1,62 @@
 {
   "name": "import-js",
   "version": "4.0.2",
-  "lockfileVersion": 1,
+  "lockfileVersion": 3,
   "requires": true,
-  "dependencies": {
-    "@babel/cli": {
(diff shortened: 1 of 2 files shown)


mikabytes:
Ah yikes.... I mistook `git fetch` for `git fetch upstream` and based this thinig on an old commit. Master here already uses sqlite3 v5. 

Closing this.

PR #590, opened 2024-01-12
Update eslint and related packages

lencioni:
I was unable to run npm install in this package because it failed a peer dependency check. To fix this I am updating some eslint packages.

Linting does not pass, but at least we can install packages.

We should really convert this repo to use prettier and TypeScript instead of flow.


mikabytes:
Strongly agree with the Prettier option! :-)

PR #598, opened 2024-01-18
594 move to prettier abandon eslint

mikabytes:
I have separated this work into 4 commits. You may want to review them separately, as otherwise prettier is making too much noise to see what is actually going on.

First make the necessary changes for prettier to start working, and eslint to stop working.

Second and third deals with four files that were improperly formatted. They didn't have the `// @flow` shebang, so tests didn't find the issues that they had. But prettier found it! So I added the shebang where I could, and removed the Flow code where there were errors.

Fourth is the actual run of prettier.

Very pretty.


trotzig:
Ignore the codeclimate status. I'm trying to get that set up again and will need some time to tweak.

trotzig:
LGTM!

mikabytes:
Hey @lencioni I'd love to have your feedback as well. Does all this look good to you?

Also a question, for this project, is it typically the reviewer who makes the merge, or the requester?

trotzig:
I would say Requester, but it's not important who does it. Go ahead and merge when you feel ready. 

(@lencioni is located in the US so probably won't check this for a few hours).

mikabytes:
Alright, thanks.

I need to switch focus to another project now, so no rush on my side getting this merged

trotzig:
Sorry if I messed up a clean merge here. Should be fairly straightforward to rebase this on the latest master. I've changed some lines that you can apply my version of. Then just run prettier on the whole thing again. 🙏

package.json
```diff
@@ -53,11 +46,6 @@
     "@babel/preset-flow": "^7.7.4",
     "@babel/preset-react": "^7.7.4",
     "babel-jest": "^24.9.0",
-    "eslint": "^8.56.0",
```
lencioni, line 56: eslint still provides plenty of value and I'd like to keep it in this project. The configuration needs to be adjusted though so it does not conflict with prettier (i.e. disable all formatting rules) and so the things it reports on are helpful to this project.

I think it would be fine to do that in a fast followup if you prefer to land this and then bring eslint back in a good state. I'd like to have eslint back on again before the next version is published.

package.json
```diff
@@ -84,13 +73,13 @@
...
-    ],
+    "setupFiles": ["./setupJest.js"],
     "testPathIgnorePatterns": [
-      ".eslintrc.js",
       "<rootDir>/build/",
       "<rootDir>/node_modules/"
     ]
+  },
+  "prettier": {
+    "semi": false
```
lencioni, line 92: semicolons are important, please change this to true.

I also would like to use single quotes, which I think is controlled by `"singleQuote": true` https://prettier.io/docs/en/options#quotes

These two changes should also help reduce the amount of diff churn on the run prettier commit.

lencioni (commented):
Thanks for working on this!

mikabytes:
> Sorry if I messed up a clean merge here.

No worries at all. I thought that would happen, which is partly why I cleanly separated the prettier run in a single commit so that I may change it later :+1:

package.json
mikabytes: I've switched it to use single quotes, and as you pointed out the diff is much smaller now. Makes it nicer to review for sure.

## As for semicolons

I'm new here, so we will go in whatever direction you say. But if I may have a few minutes of your time, I'd like to argue my point:

They are important where necessary, and useless everywhere else. It comes down to personal preference. I prefer not to see useless things, but I can endure.

- Prettier never changes how the code _works_
- Prettier only removes them where they are unnecessary (unimportant)
- Prettier does not remove them where they serve a purpose
- It will still add semis where there are common pitfalls for humans
- It will wrap lines where there might be ambiguity, making it obvious to the human how that code will run
- You can still write semicolons as you code, if you wish, then let prettier do its thing
- If we have semis ON, I can still choose to code without them, then let prettier do its thing

### Example 1

```
a = ["a"]
(a + b)
```

becomes...

```
a = ["a"](a + b)
```

_Uh oh, that turned into a function call, I see what I made wrong there..._

### Example 2

```
a = ["a"];
(a + b) * 3;
```

becomes...

```
a = ["a"]
;(a + b) * 3
```

_What? A semicolon was inserted before my row? Maybe I did something odd here..._

To add some weight to my argument; Since I changed my mind about them back in 2017 I've had some 30 people working without semicolons and we've never had a bug resulting from omitting them while Prettier does its magic. I can count on one hand where I made an assumption about ASI that proved wrong, and in each case it was obvious.

package.json
mikabytes: Oh, I must've missed something. What are we using ESlint for other than syntax?

I can revert that commit, or we can make a separate PR after this one. Just need to understand what to keep before I start working on it.

trotzig: It looks like the .eslintrc.js config extends `eslin-config-airbnb-base` which is full of useful checks. The main ones I like are the one that prevents leaving unused variables/imports behind (no-unused-vars) and the one that prevents using undefined variables (no-undef).

mikabytes: Ah, I see.

I'm feeling a little hesitant to take this on. Perhaps one of you would be up for reintroducing ESlint? Can do this in this PR if you want, or a follow-up, as you wish. Else I will try to work out some time to get reacquainted with ESlint and all its configuration possibilities.

package.json
lencioni: Thank you for the explanation. I've given this a lot of thought over the weekend.

As you pointed out, we can choose to type the semicolon or not type it, and prettier will add it or remove it based on the configuration, which is very nice.

While I understand your point about not wanting to see things that are useless, I'd like to offer the following thoughts:

- I find the semicolons to make the code easier to read. I think of it as punctuation at the end of a sentence, which helps break things up for my eyes. Agreed that this is a personal preference and that others may find the opposite to be true.
- You mentioned that there are a few cases where semicolons are required in order for the code to be parsed and executed correctly. I tend to favor rules that are biased towards simplicity. In this case, since semicolons are sometimes required it is a simpler rule to say "semicolons should always be used" vs "semicolons should never be used except when they are required to avoid bugs". Though it is nice that the tooling can handle this for us and in practice we should not need to think about it, I still favor using the simpler rule.
- This project has historically required semicolons, and there is some value in maintaining that inertia and not making this change. Granted, that's not a great argument for making changes that are important, but as you said this comes down to personal preference and is not a critical change (vs. e.g. fixing a bug or adding a useful feature).
- By sticking with the existing format of requiring semicolons, we will reduce git diff churn. This will make reviewing this PR easier (minor benefit) and will reduce friction when using log diving tools like `git blame` (medium benefit).

At minimum, if you want to keep pursuing this, it would be nice to move the controversial bits to a separate PR so we can land the rest of this one and move forward with other work that builds on top of it.

package.json
lencioni: I can re-introduce it in a follow up.

mikabytes: Great!

mikabytes:
Merged this as we agreed to keep the semis and move to single quotes and there were no other concerns. But I realized maybe @lencioni wanted to review an extra time? If so, sorry! We can revert.

PR #599, opened 2024-01-18
Replace Travis with GitHub Actions, support node 18 + 20

### LINKED ISSUE #596: Abandon old node versions, update test matrix
We currently test these node versions in our Travis tests: 
```
node_js:
  - "8"
  - "10"
  - "12"
  - "13"
```

I suggest we update this to use v 18-20 instead (LTS + current).

trotzig:
The Travis setup is broken, and I think people are more comfortable with GitHub actions these days, so I'm switching over. In the process, I updated the Node version matrix to include 18 and 20 (LTS + current).

Fixes #596


mikabytes (approved):
Sweetness!

PR #609, opened 2024-01-24
No more implicit React identifiers

mikabytes:
Much inspiration from #580, thanks @madox2!

This PR never includes implicit React imports, unlike that PR. See issue for more information.

@trotzig Do you have some React projects where you could try this out in a sharp real-life environment? I don't do React, so I can't be too sure. All the tests are passing fine.

### DIFF
--- a/lib/__tests__/Importer-test.js
+++ b/lib/__tests__/Importer-test.js
@@ -2705,8 +2705,6 @@ var b = bar + foo;
         packageDependencies = ['react'];
         existingFiles = ['bar/foo.jsx'];
         text = `
-import React from 'react';
-
 const foo = <Foo/>;
         `.trim();
       });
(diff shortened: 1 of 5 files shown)


mikabytes:
I verified this today running the steps below. I think this should be good to merge, but I'm not 100% sure as I haven't worked with React since some 8 years ago.

Steps to test:

```
mkdir test ; cd test

npm init

echo "const a = <Test />" > test.js

importjs fix test.js

cat test.js
```

Result before PR:

```json
{"messages":["Imported React from 'react'"],"fileContent":"import React from 'react'\n\nconst a = <Test />;\n","unresolvedImports":{}}
```

Result after PR:

```json
{"messages":[],"fileContent":"const a = <Test />\n","unresolvedImports":{}}
```

mikabytes:
No worries @trotzig, glad to hear you're keeping those priorities straight :+1:

Issue #571, opened 2021-05-20
Unable to import files when using smart pipeline operator


LandonSchropp:
I have my project set up to use the smart pipeline operator via the `babel/plugin-proposal-pipeline-operator` plugin. When I'm in a file using this operator, importjs fails to import my modules with an error.

Here's an example of a file:

```
export function mapDouble(array) {
  return array |> _.map(#, x => x * 2);
}
```

When I run `importjs fix source/javascript/example.js`, I get the following error:

```
{
  "messages": [
    "SyntaxError: Primary Topic Reference found but pipelineOperator not passed 'smart' for 'proposal' option. (2:24)"
  ],
  "fileContent": "export function mapDouble(array) {\n  return array |> _.map(#, x => x * 2);\n}\n",
  "unresolvedImports": {}
}
```

This code works fine inside my app, but not with importjs. Here's my `.babelrc`:

```
{
  "presets": [ "babel-preset-react-app" ],
  "plugins": [
    [ "@babel/plugin-proposal-pipeline-operator", { "proposal": "smart" } ]
  ]
}
```

Thanks!

mikabytes:
@lencioni @trotzig how are you feeling about adding support to Stage 2+ proposals to ImportJS?

I believe it would be an easy implementation (perhaps just a single line somewhere around https://github.com/Galooshi/import-js/blob/master/lib/parse.js#L12

Can you see any drawbacks to going down that path?

trotzig:
I'm okay with experimental stuff in general. We basically just need import-js to be able to parse and understand a file as much as possible. Unless the proposal is breaking some old way of writing code, I think we can add support.

mikabytes:
Just found we already support `minimal` pipeline operator. However, `minimal` has since been [deprecated](https://babeljs.io/docs/babel-plugin-proposal-pipeline-operator#usage).

The only non-deprecated proposal is `hack` (not to confuse with being a hack). I'm in favor of moving to that, however, it comes with some drawbacks.

## 1. This is a breaking change

Current ImportJS supports code like this:

```
        "some value" |> console.log;
```

After changing to `hack` it would have to be written

```
        "some value" |> console.log(%);
```

## 2. It requires specifying a topic token

The proposal suggests using `%` as the topic token but also makes clear that it is not a final decision. Babel provides the `topicToken` field to let the user configure it.

[...]

PR #611, opened 2024-01-26
Provide the option to override parser plugins in configuration

mikabytes:
Related to #571

This PR provides a new configuration key `parserPlugins` that allows the end-user to override or extend what babel parser plugins to use. See [README](https://github.com/Galooshi/import-js/pull/611/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R582) for example use.

This implementation uses a (nasty) singleton in parse.js to achieve the goal. While I'm not too happy about that, I have reasons;

I spent a long time trying to implement this by passing the values around from Configuration.js to its endpoint, parse.js. However, I ran into trouble doing so. As it turns out a lot of code is calling parse.js indirectly, such as `findExports` and even `meteorEnvironment` and `Configuration` itself. I tried to find all the cases but in the end I had hundreds of lines of code changed and 96 failed tests. Weird circular dependencies started showing up as well. In the end, I gave up on that strategy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -573,6 +573,54 @@ module.exports = {
 };
 ```
 
+### `parserPlugins`
+
+ImportJS defaults to a reasonable compromise for what syntax to support but can be overridden (replaced) in configuration. The latest defaults can be found [here](lib/parse.js#L3)
+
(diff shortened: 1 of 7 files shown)


trotzig (approved):
Thanks!

PR #612, opened 2024-01-28
Add configuration validation and refactor handling of unknown identifiers

mikabytes:
- Add validation using Ajv and JSON Schema 
- Moved the "unknown identifiers" management there
- Moved the default schema there

Our configuration schema is big and complex. This is mostly due to every property having the ability to also be a function (dynamic configuration). 

Several properties don't belong in configuration. We might want to look into extracting them at a later point. Examples are `cacheLocation`, `coreModules`.. These are also undocumented. They bring in a hairball of dependencies that don't make sense for a schema to deal with. It would clean things up a bit to refactor them out.

This implementation will return messages on bad configuration and delete the offending keys. Alternatively, we could crash, but seeing as we have earlier ignored them this seems to be the most reasonable way to go.

Changed one test using Set to use Array instead, as I couldn't validate Sets using `instanceof`. Not sure why. But it works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -534,7 +534,6 @@ Available plugins are over at [Babel: Plugins List](https://babeljs.io/docs/plug
 
 #### Example: Remove all preconfigured defaults
 
-```javascript
 parserPlugins: []
 ```
 
@@ -571,7 +570,7 @@ parserPlugins: [
       proposal: 'hack',
     },
   ],
-]
+];
 ```
 
 ### `sortImports`
--- a/README.md
+++ b/README.md
@@ -534,6 +534,7 @@ Available plugins are over at [Babel: Plugins List](https://babeljs.io/docs/plug
 
 #### Example: Remove all preconfigured defaults
 
+```javascript
 parserPlugins: []
 ```
 
--- a/lib/Configuration.js
+++ b/lib/Configuration.js
@@ -27,8 +27,6 @@ const ENVIRONMENTS = {
 };
 
 function checkConfiguration(config: Object): Array<string> {
-  const messages = [];
-
   const result = validate(config);
 
   return result.messages;
--- a/lib/Configuration.js
+++ b/lib/Configuration.js
@@ -1,119 +1,23 @@
 // @flow
 
-import crypto from 'crypto';
 import os from 'os';
 import path from 'path';
 import has from 'lodash/has';
 
-import globals from 'globals';
 import semver from 'semver';
 
 import FileUtils from './FileUtils';
 import JsModule from './JsModule';
-import findPackageDependencies from './findPackageDependencies';
 import meteorEnvironment from './environments/meteorEnvironment';
 import nodeEnvironment from './environments/nodeEnvironment';
 import normalizePath from './normalizePath';
 import version from './version';
-import { DEFAULT_PARSER_PLUGINS } from './parse.js';
+import { validate, getDefaultConfig } from './configurationSchema.js';
 
 const JSON_CONFIG_FILE = '.importjs.json';
 const JS_CONFIG_FILES = ['.importjs.js', '.importjs.cjs', '.importjs.mjs'];
 
-function findGlobalsFromEnvironments(
-  environments: Array<string>,
-): Array<string> {
-  const result = Object.keys(globals.builtin);
-
-  environments.forEach((environment: string) => {
-    const envGlobals = globals[environment];
-    if (!envGlobals) {
-      return;
-    }
-    result.push(...Object.keys(envGlobals));
-  });
-  return result;
-}
-
-const DEFAULT_CONFIG = {
-  aliases: {},
-  declarationKeyword: 'import',
-  cacheLocation: ({ config }: Object): string => {
-    const hash = crypto
-      .createHash('md5')
-      .update(`${config.workingDirectory}-v4`)
-      .digest('hex');
-    return path.join(os.tmpdir(), `import-js-${hash}.db`);
-  },
-  coreModules: [],
-  namedExports: {},
-  environments: [],
-  excludes: [],
-  globals: ({ config }: Object): Array<string> =>
-    findGlobalsFromEnvironments(config.get('environments')),
-  groupImports: true,
-  ignorePackagePrefixes: [],
-  importDevDependencies: false,
-  importFunction: 'require',
-  importStatementFormatter: ({ importStatement }: Object): string =>
-    importStatement,
-  logLevel: 'info',
-  maxLineLength: 80,
-  minimumVersion: '0.0.0',
-  moduleNameFormatter: ({ moduleName }: Object): string => moduleName,
-  moduleSideEffectImports: (): Array<string> => [],
-  sortImports: true,
-  emptyLineBetweenGroups: true,
-  stripFileExtensions: ['.js', '.jsx', '.ts', '.tsx'],
-  danglingCommas: true,
-  tab: '  ',
-  useRelativePaths: true,
-  packageDependencies: ({ config }: Object): Set<string> =>
-    findPackageDependencies(
-      config.workingDirectory,
-      config.get('importDevDependencies'),
-    ),
-  // Default configuration options, and options inherited from environment
-  // configuration are overridden if they appear in user config. Some options,
-  // however, get merged with the parent configuration. This list specifies which
-  // ones are merged.
-  mergableOptions: {
-    aliases: true,
-    coreModules: true,
-    namedExports: true,
-    globals: true,
-  },
-  parserPlugins: DEFAULT_PARSER_PLUGINS,
-};
-
-const KNOWN_CONFIGURATION_OPTIONS = [
-  'aliases',
-  'cacheLocation',
-  'coreModules',
-  'declarationKeyword',
-  'environments',
-  'excludes',
-  'globals',
-  'groupImports',
-  'ignorePackagePrefixes',
-  'importDevDependencies',
-  'importFunction',
-  'importStatementFormatter',
-  'logLevel',
-  'maxLineLength',
-  'minimumVersion',
-  'moduleNameFormatter',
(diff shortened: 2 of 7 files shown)


mikabytes:
@lencioni If it's convenient for you, would you mind having a look at this code? 

Or, if you're pressed for time, perhaps I should go ahead and merge it. Should be fairly safe.

Thanks

README.md
```diff
@@ -534,7 +534,6 @@ Available plugins are over at [Babel: Plugins List](https://babeljs.io/docs/plug
 
 #### Example: Remove all preconfigured defaults
 
-```javascript
```
lencioni, line 537: Why remove this?

lib/Configuration.js
```diff
@@ -122,16 +26,12 @@ const ENVIRONMENTS = {
   meteor: meteorEnvironment,
 };
 
-function checkForUnknownConfiguration(config: Object): Array<string> {
+function checkConfiguration(config: Object): Array<string> {
   const messages = [];
```
lencioni, line 30: This seems to be unused now

lib/configurationSchema.js
```diff
@@ -0,0 +1,438 @@
...
+          config.get('importDevDependencies')
+        ),
+      anyOf: [
+        {
+          instanceof: 'Function',
+        },
+        {
+          type: 'array',
+          items: {
+            type: `string`,
```
lencioni, line 288: Should probably change the back ticks to single quotes

lib/configurationSchema.js
```diff
@@ -0,0 +1,438 @@
...
+
+function findGlobalsFromEnvironments(environments) {
+  const result = Object.keys(globals.builtin);
+
+  environments.forEach((environment) => {
+    const envGlobals = globals[environment];
+    if (!envGlobals) {
+      return;
+    }
+    result.push(...Object.keys(envGlobals));
```
lencioni, line 378: Do we want this to remove duplicates?

lib/configurationSchema.js
```diff
@@ -0,0 +1,438 @@
...
+  });
+  return result;
+}
+
+export function getDefaultConfig() {
+  return Object.entries(SCHEMA.properties).reduce((acc, [k, v]) => {
+    if (typeof v !== `object` || v === null) {
+      throw new Error(`Expected schema key '${k}' to be an object`);
```
lencioni, line 386: Instead of saying "key" I think it might make more sense to say "value at" or something similar. It might also be nice to log typeof v.

lib/configurationSchema.js
```diff
@@ -0,0 +1,438 @@
...
+    }
+    if (v.hasOwnProperty('default')) {
```
lencioni, line 388: What do you think about using Object.hasOwn here instead?

lib/configurationSchema.js
```diff
@@ -0,0 +1,438 @@
...
+  const validate = ajv.compile(SCHEMA);
+
+  if (!validate(data)) {
+    return {
+      error: true,
+      messages: validate.errors
+        .map((err) => {
+          // report unknown identifiers the same as we have done in the past
+          // so we can showcase that the move to Ajv didn't change previous
+          // behaviors
```
lencioni, line 408: I appreciate your attention to detail but I do wonder how important is it to preserve this exact behavior?

lib/configurationSchema.js
```diff
@@ -0,0 +1,438 @@
...
+          if (err.message === 'must NOT have additional properties') {
+            // remove the extraneous identifier so we avoid later errors in code
+            // instancePath is empty if root key
+            const rootKey =
+              err.instancePath.split('/')[1] || err.params.additionalProperty;
+            delete data[rootKey];
+
+            return (
+              'Unknown configuration: `' +
+              (err.instancePath ? err.instancePath + `.` : ``) +
```
lencioni, line 418: Use quotes here too

lib/configurationSchema.js
```diff
@@ -0,0 +1,438 @@
...
+            const rootKey = err.instancePath.split('/')[1];
+            delete data[rootKey];
+
+            console.log(err);
+
+            return 'Invalid configuration: `' + err.instancePath.slice(1) + '`';
+          }
+        })
+        .filter((val, index, a) => a.indexOf(val) === index)
+        .filter((it) => !!it),
```
lencioni, line 433: This could be: .filter(Boolean)

lib/configurationSchema.js
```diff
@@ -0,0 +1,438 @@
...
+              (err.instancePath ? err.instancePath + `.` : ``) +
+              err.params.additionalProperty +
+              '`'
+            );
+          } else {
+            // remove the failing identifier so we avoid later errors in code
+            const rootKey = err.instancePath.split('/')[1];
+            delete data[rootKey];
+
+            console.log(err);
```
lencioni, line 427: Should this use console.error?

lencioni (approved):

lencioni:
Thanks for doing this! I reviewed on a phone so sorry about any weird formatting

lib/Configuration.js
mikabytes: Woah, good catch.

I sorted all the configuration options in another PR. This caused a merge conflict here, and I must've done somthing wrong when resolving it.

I have re-added it now.

lib/configurationSchema.js
mikabytes: Ah, old habits die hard.

In a previous company, we resolved the "single quotes vs double quotes" debate with "always use backticks".

I've changed it to single quotes now.

lib/configurationSchema.js
mikabytes: This function was moved from [configuration.js](https://github.com/Galooshi/import-js/pull/612/files/ae686c28c6cd7d39c63c44368f56caf3f68cdb25#diff-94c307851d4ad36c747a9f883703a4c1f964b82c23320b47899c4e543fd9a8f9L23) because the defaults depend on it.

I would suggest that changes to it be made in another PR, trying to keep this as pure "configuration schema validation" as possible. What do you think?

lib/configurationSchema.js
mikabytes: Changed it to 
```
`Expected schema key '${k}' to be an object, but it was of type '${typeof v}'. Got: ${v}`
```

This error will only happen when we change the defaults in code, so it should be fairly obvious what key has changed, but it's always nice to provide as much information as possible.

lib/configurationSchema.js
mikabytes: Either one is okay by me. Possibly one downside of using Object.hasOwn is that we would be losing support for Node <16.8. That might be okay, it's a bit over two years old. I'd probably want a higher benefit/disadvantage ratio.

Maybe we better deal with that the day we choose to use `hasOwnProperty` as a key for our configuration :smile:

lib/configurationSchema.js
mikabytes: Not important at all :smile: But it helps reduce the scope of this PR and lowers my anxiety of having accidentally introduced some unwanted side-effect.

I was thinking of improving error messages in an upcoming PR. 

Faulty identifiers / bad configuration would be very nice to have the information of what the offending value was, and how to fix it. Ajv gives me a lot of duplicate error messages, as well as false positives when using `anyOf`, so I think that stuff will need a bit of care anyway.

lib/configurationSchema.js
mikabytes: Yep

lib/configurationSchema.js
mikabytes: I would have preferred something that conveys my intention better, "isTruthy" or something, but Boolean will have to do :+1:

lib/configurationSchema.js
mikabytes: Ah, I left that in there by mistake. I've removed it now.

Logging things in low-level modules makes the tests very chatty, and I think that's anyway not the right place to communicate this information. We already return the error messages, so it is the responsibility of the caller function to do so, if needed.

mikabytes:
@lencioni Thanks for your feedback, much appreciated!

I put some comments of my own. I'll wait for you to review them, and the changes, before merging this.

lib/configurationSchema.js
lencioni: Makes sense

lib/configurationSchema.js
lencioni: Better but still the wording is a bit off I think. We are talking about the value at that key, not the key itself, but the way it is rded could be understood as referring to the type of the key itself and not its value.

lib/configurationSchema.js
lencioni: Yeah, or could use Object.prototype.hasOwnProperty.call

lib/configurationSchema.js
lencioni: Improving error messages sounds great to me. I wonder if anything could be added to the schema that would help with that, or if ajv offers any options to tweak the output it generates out of the box. Or maybe someone else has put out a package that makes ajv errors really nice?

lib/Configuration.js
lencioni: I think you responded to the wrong comment here. This array is still unused.

mikabytes: Ah yes. Apologies. I responded to the wrong comment.

The unused variable 'messages' has now been removed.

lib/configurationSchema.js
mikabytes: Good ideas. There must be. I will look into it.

I was also thinking we might move documentation into the JSON schema and auto-generate it. I came across [jsonschema2md](https://github.com/adobe/jsonschema2md) that might be interesting for this. But that's yet another PR.

mikabytes: Started #615 for invalid configuration error messages and #616 for auto-generating documentation

PR #620, opened 2024-02-22
Disable codeclimate when env var is missing

### LINKED ISSUE #619: Missing CODECLIMATE_REPO_TOKEN for Github Actions
Just noticed in #618 that the tests are failing. It's mentioning something about this env variable. @trotzig do you have a few minutes to check it out? Thanks.

trotzig:
When outside contributors, e.g. dependabot, pushes a PR, the CODECLIMATE_REPO_TOKEN env variable is not present (it's a secret defined on the github account). We can ignore the codeclimate report in these cases to let tests still pass.

An alternative option would be to make the codeclimate secret available to everyone, but that's not very safe.

Fixes #619


mikabytes:
Great, thanks for taking the time to work on this. I know you're busy :+1:

Issue #608, opened 2024-01-24
Reintroduce some ESlint rules


mikabytes:
- `no-unused-vars`
- no useless returns

... probably several others.

These were removed as part of #598

Issue #593, opened 2024-01-18
Move away from Flow?


mikabytes:
~~This project uses `.js` extension for files that include TypeScript. This confuses some editors, mine included. Perhaps we could change them to `.ts`?~~ I was confused :-)

Let's discuss whether or not to move away from Flow, and what to move towards.

trotzig:
I think they contain Flowscript, right? I don't use much TypeScript myself but I would be okay with switching over. If I was the only maintainer I would probably just use straight javascript but I can see the benefits of types.

mikabytes:
Oh, I thought Flow was a layer on top of TypeScript. Full disclosure; I've never worked actively in either. My ignorance is showing :-)

If I was the only maintainer I would also go for straight vanilla JavaScript. Because that's what I know best and am used to.

So maybe there's a discussion to be had about Flow. I think I saw @lencioni mention in a comment he might want to move away from it. 

What other people would be interested in this conversation?

trotzig:
I can't speak for @lencioni but I I'm pretty sure he would suggest we use TypeScript. But then again, I don't think he cares too much since he's not using the project himself. 

I would be in favor of using plain old javascript, for a few reasons: 
- More people can contribute
- No need for a build step (simplifies things)

I would need some time to ramp up with TypeScript, updating my dev environment, reading up on best practices, etc. And I'd prefer not to do that right now. But that's my laziness talking, I'm okay getting my hands dirty with some types as well. 🙂

mikabytes:
While I also have some of that "getting up to speed" on TypeScript, I think the biggest selling point of using plain old JavaScript is that it is the de-facto standard. It is the subset that everybody knows. I also appreciate shedding some weight considering tooling (build steps etc)

Meanwhile, a strong selling point of typed languages is to catch bugs early, another is to add IDE assistance. I use tests for finding bugs, and as a Vim user, I don't need all that sugar.

That being said, if TypeScript is the way to go then I'm fine with that. Probably won't be me refactoring the codebase though ;)

lencioni:
I think we should move this project to TypeScript. Beyond what's already been said, there are a few more reasons:

1. We already have Flow type annotations, so the conversion to TypeScript should be a little easier from the current state.
2. Getting your feet wet with TypeScript would be easiest in a smaller project, like this one, so I think it is the ideal opportunity for both of you.

> I use tests for finding bugs

[...]

PR #623, opened 2024-06-07
Remove flow type annotations, set up ESLint

lencioni:
Related issues:

- https://github.com/Galooshi/import-js/issues/593
- https://github.com/Galooshi/import-js/issues/608

I decided to remove flow type annotations with this change because I didn't want to bother with setting up the new ESLint to understand the flow syntax. I also think that it might be easier to adopt TypeScript from this state than it would be to attempt to port over the flow types.

For now, I am just rolling with the default recommended ESLint configuration that they include with the base package, to get a decent baseline of ESLint rules enabled.

This is probably best reviewed commit-by-commit instead of all at once.


mikabytes (approved):
Good to see some progress on this. Thank you!

All looks good to me. 

One idea; maybe we could include eslint in `npm test`. That way we cover tests, prettier, and eslint all in one command. Less likely to miss bugs before setting up a pull request.

trotzig (approved):
Looks good to me! Thanks @lencioni!

